PR #3, opened 2020-10-23
Allow using collier provided externally

andrey-popov:
`collier` (version 1.2.4) is [available](http://lcginfo.cern.ch/pkg/collier/) in LCG software stacks&dagger;, so there is no need to build it if this environment is used. This commit will skip the build of `collier` if environmental variable `COLLIER_ROOT_DIR` is set. In that case the rest of `JHUGenMELA` will assume that the interface modules and the shared library of `collier` are located within directories `$COLLIER_ROOT_DIR/include` and `$COLLIER_ROOT_DIR/lib` respectively. If this variable is not set, `collier` will be built in place as before.

Skipping compilation of `collier` reduces the total build time (run in one thread) from 470 to 310 s.

I checked that the package compiles both when `COLLIER_ROOT_DIR` is set and not set. After building with `COLLIER_ROOT_DIR`, all tests in `test/testME_all.py` succeed.

&dagger;This works with `LCG_97a` and later. `collier` was included also in earlier releases, but they missed `collier.mod` interface module.

### DIFF
--- a/MELA/COLLIER/setup.sh
+++ b/MELA/COLLIER/setup.sh
@@ -35,17 +35,24 @@ if [[ $# > 0 ]] && [[ "$1" == *"clean"* ]]; then
 else
 
   if [[ ! -f "../data/$SCRAM_ARCH/$libname" ]]; then
-    rm -rf $tmpdir
-    rm -f $tarname
-    wget --no-check-certificate $tarweb
-    mkdir $tmpdir
-    tar -xvzf $tarname -C $tmpdir
-    rm -f $tarname
-    mv $tmpdir"/"$pkgdir"/src/"* ./
-    rm -rf $tmpdir
+    if [[ -z "${COLLIER_ROOT_DIR+x}" ]]; then
+      # Build a copy of collier in place
+      rm -rf $tmpdir
+      rm -f $tarname
+      wget --no-check-certificate $tarweb
+      mkdir $tmpdir
+      tar -xvzf $tarname -C $tmpdir
+      rm -f $tarname
+      mv $tmpdir"/"$pkgdir"/src/"* ./
+      rm -rf $tmpdir
 
-    make $@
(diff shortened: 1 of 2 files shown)


usarica:
Hi @andrey-popov 
I think it is better to keep the build customized because Fortran interface to C++ might depend on what flags are used. This is why we have our own makefile to control the compilation a bit more strictly than to rely on externally-provided libraries (just like MCFM).

andrey-popov:
But if `JHUGenMELA` compiles successfully, doesn't this mean that there are no problems in the interface with Fortran?

usarica:
Right now, it might (and the main reason is probably because there is no C-interface done yet to COLLIER itself; it is in the plans though), but if the flags change, causing changes in the expressions of the functions in the library (e.g. when you run 'nm'), then linking via extern statements in the interface would need to be modified for how many underscores are needed (and where), logical -> bool vs int C correspondence etc. (see for example interface/TMCFM.hh vs interface/TModJHUGen.hh). Using an externally-compiled library would not be sustainable for this reason; this is why we choose the flags to fix the interface details explicitly.

usarica:
Sorry, meant to point to interface/TModJHUGenMELA.hh, which illustrates the logical->int type of interface need, not interface/TModJHUGen.hh.

usarica:
BTW, I think it is good that you point there is v1.2.4 available. We should probably test that for compatible results and upgrade to it.

andrey-popov:
Are you saying that Fortran doesn't provide a stable interface with C? That's unfortunate… Feel free to close my PR then.

usarica:
There are certain basic rules, but the signature of the functions in the libraries might change depending on the flags. This is where dependency on the flags come in. There are also differences between F77 vs F90+ interfacing details as you might have noticed (necessarily because of how Fortran modules operate wrt plan Fortran functions), but that is more stable since the Fortran code written in F77 or F90 rarely changes its language type (not that it doesn't ever happen though).

PR #4, opened 2020-11-09
Delete created symbolic links in Mela destructor

mostafa1993:


usarica:
@mostafa1993 Thanks for the PR. However, I see that the filesystem header is only available for C++17 while our code must support C++11. Would you please implement a C++11-compatible set of remove statements?

usarica:
> @mostafa1993 Thanks for the PR. However, I see that the filesystem header is only available for C++17 while our code must support C++11. Would you please implement a C++11-compatible set of remove statements?

Basically, by using unistd.h::remove.

mostafa1993:
I see. I just got confused how to update my PR so I will do it in a new PR.

PR #5, opened 2020-11-09
Enable deleting created symbolic links

mostafa1993:


usarica:
@mostafa1993 Thanks for the PR. Putting the rm statement in the destructor is not necessarily an ideal one. Consider the scenario where there are more than one Mela objects, each one being deleted separately. Then deletion of one object could disrupt the operation of the other objects. My suggestion would be to move these statements to a static function (static void Mela::cleanLinkedFiles()). Then the user can call it in the destructor of their event looper.

mostafa1993:
> @mostafa1993 Thanks for the PR. Putting the rm statement in the destructor is not necessarily an ideal one. Consider the scenario where there are more than one Mela objects, each one being deleted separately. Then deletion of one object could disrupt the operation of the other objects. My suggestion would be to move these statements to a static function (static void Mela::cleanLinkedFiles()). Then the user can call it in the destructor of their event looper.

Done :)

usarica:
@mostafa1993 Thanks for making the change quickly.

@TOPAZdevelop @gritsan @hroskes Would someone please approve the PR? It adds a clean-up functionality without disrupting core functions.

usarica (approved):

usarica (approved):
The added changes are necessary for JHUGenMELA to function in later versions of HTCondor and glidein migrations. They need to be approved for JHUGenMELA to function on remote machines.

TOPAZdevelop (approved):

PR #6, opened 2021-03-11
Version Change 

jdavis36:
Change the JHUGen version to pass the CircleCI test.

### DIFF
--- a/MELA/fortran/mod_Parameters.F90
+++ b/MELA/fortran/mod_Parameters.F90
@@ -3,7 +3,7 @@ implicit none
 save
 !
 !
-character(len=*),parameter :: JHUGen_Version="v7.4.0"
+character(len=*),parameter :: JHUGen_Version="v7.4.1"
 !
 !
 !=====================================================


usarica:
I can approve, but the JHUGenMELA commit id will need to be updated afterward. @hroskes @gritsan @TOPAZdevelop we would need one more approval.

usarica (approved):

gritsan (approved):

hroskes:
Hi Ulascan, I don't recommend squash merging for MELA.  It interacts in a complicated way with submodules.  I was just helping Jeff debug but it would be easier to just stick to normal merging.

usarica:
Hi Heshy,

How would it interact in a complicated way? Is it not just a matter of pointing to the correct commit id in JHUGen/JHUGenMELA? One should just refer to the commit id in the repository, not their fork.

usarica:
Besides, I would like the commit history be as linear as possible. A merge commit followed by a merge commit is not something I would prefer to see in the history after merges.

hroskes:
I understand why you wanted to squash, and I agree it's worth avoiding merge commits when they're not necessary.  But orphaned references to submodules cause bigger problems than complicated histories do.

But if there was an older test version of the generator from before the MELA PR was merged, and it refers to either the commit used for the PR or an earlier commit in the branch that was later squashed, then it doesn't automatically get fetched when you check out MELA and you have to do some hacks to get it.  Even if eventually it gets updated, having an orphaned submodule reference in the history has caused me problems in the past.

I once had an orphaned reference to a commit that was force pushed and was gone from any history, which is an even bigger nightmare.  If it's in a PR, that's better in the sense that it can be fetched, and that's how I helped Jeff fix it for circleci.  But please trust me that doing this too often is going to cause problems.

PR #7, opened 2021-03-18
tHW added

tmartini:
in order for CircleCI tests "misc" to succeed

### DIFF
--- a/MELA/fortran/mod_Misc.F90
+++ b/MELA/fortran/mod_Misc.F90
@@ -1548,7 +1548,7 @@ character(len=*) :: title
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D86 (2012) 095031;  arXiv:1208.4018  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D89 (2014) 035007;  arXiv:1309.4819  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D94 (2016) 055023;  arXiv:1606.03107 [hep-ph].", linelength)
-    write(TheUnit, *) " ", CenterWithStars("                              arXiv:2002.09888 [hep-ph].", linelength)
+    write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D102 (2020) 056022; arXiv:2002.09888 [hep-ph].", linelength)
     write(TheUnit, *) " ", CenterWithStars("", linelength)
     write(TheUnit, *) " ", repeat("*", linelength)
     write(TheUnit, *) " "
--- a/MELA/fortran/mod_Misc.F90
+++ b/MELA/fortran/mod_Misc.F90
@@ -1548,7 +1548,7 @@ character(len=*) :: title
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D86 (2012) 095031;  arXiv:1208.4018  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D89 (2014) 035007;  arXiv:1309.4819  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D94 (2016) 055023;  arXiv:1606.03107 [hep-ph].", linelength)
-    write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D102 (2020) 056022; arXiv:2002.09888 [hep-ph].", linelength)
+    write(TheUnit, *) " ", CenterWithStars("                              arXiv:2002.09888 [hep-ph].", linelength)
     write(TheUnit, *) " ", CenterWithStars("", linelength)
     write(TheUnit, *) " ", repeat("*", linelength)
     write(TheUnit, *) " "
@@ -1585,6 +1585,8 @@ function CalculatesXsec(Process)
   elseif (Process.eq.80 .or. Process.eq.90) then
     CalculatesXsec=.false.
   elseif (Process.ge.110 .and. Process.le.114) then
+    CalculatesXsec=.true.
+  elseif (Process.ge.115 .and. Process.le.117) then
     CalculatesXsec=.true.
   else
     print *, "Unknown process in CalculatesXsec", process, "; setting CalculatesXsec=false."
--- a/MELA/fortran/mod_THiggs.F90
+++ b/MELA/fortran/mod_THiggs.F90
@@ -464,10 +464,17 @@ integer, parameter :: inLeft=1,inRight=2,Hbos=3,t=4, bout=5, bdk=6,W=7,lep=8,nu=
 RETURN
 END SUBROUTINE
 
-
+!amplitude squared for partonic reaction 
+!
+! g(iL) + b(iR) -> t(t) + H(H) + Wm(w) 
+!              (-> b(bt) + Wp(wt) + Wm(w)
+!               -> b(bt) + f1~(lept) + f'1(nut) + f'2(lepW) + f2~(nuW))
+!
+!with the fermion-antifermion pairs f'1/f1~ and f'2/f2~ 
+!from leptonic or hadronic decays of the Wp/Wm   
+!
 SUBROUTINE EvalAmp_GB_TWMH(MomExt,LO_Res_Unpol)
 use ModMisc
-use ModTTBHiggs
 use ModParameters
 use ModTopDecay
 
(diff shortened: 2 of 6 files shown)


jdavis36 (approved):
These changes are required to match the JHUGen Fortran files that now include the tWH process.
Changes were already approved for JHUGen therefore I approve these changes.

jdavis36 (approved):

gritsan (approved):

MELA/fortran/mod_Misc.F90
```diff
@@ -1548,7 +1548,7 @@ SUBROUTINE PrintLogo(TheUnit, title)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D86 (2012) 095031;  arXiv:1208.4018  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D89 (2014) 035007;  arXiv:1309.4819  [hep-ph],", linelength)
     write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D94 (2016) 055023;  arXiv:1606.03107 [hep-ph].", linelength)
-    write(TheUnit, *) " ", CenterWithStars("Phys.Rev. D102 (2020) 056022; arXiv:2002.09888 [hep-ph].", linelength)
+    write(TheUnit, *) " ", CenterWithStars("                              arXiv:2002.09888 [hep-ph].", linelength)
```
usarica, line 1551: The change on this line should be reverted.

MELA/fortran/mod_TopDecay.F90
```diff
@@ -100,6 +100,52 @@ SUBROUTINE TopDecay(Flavor,Mom,Spinor,TopHel)
 
 
 
+
+SUBROUTINE WDecay(Charge,Mom,WCurr)
```
usarica, line 104: Do you think you could describe how this function is used?

usarica (changes requested):

MELA/fortran/mod_TopDecay.F90
tmartini: this function is for the decay of a W boson with polarization vector WCurr and momentum Mom.
It is needed for the spin-correlated decay of the W boson in tHW production (cf. TopDecay(Flavor,Mom,Spinor,TopHel))

tmartini: to clarify: this function is basically the W-decay part taken from the TopDecay subroutine, thus it appears in mod_TopDecay. For tHW both the top decay and the W decay are needed in terms of the spinors/polarization vectors used in the stable production.

tmartini:
is there anything else which hinders the approval?

gritsan (approved):

MELA/fortran/mod_THiggs.F90
```diff
@@ -437,6 +465,149 @@ SUBROUTINE EvalAmp_QQB_TBARHB(MomExt,LO_Res_Unpol)
...
+use ModTTBHiggs
+use ModParameters
+use ModTopDecay
+
+implicit none
+real(8) :: MomExt(1:4,1:11),LO_Res_UnPol(-6:6,-6:6)
+complex(8) :: LOAmpab=0d0, LOAmpba=0d0
+
+complex(8) :: UBbD(4),UBbD2(4),Eg(4),Eg2(4),UBtD(4),ECw(4)
+integer, parameter :: iL=1,iR=2, H=3, t=4, w=5, bt=6,Wt=7,lept=8,nut=9, lepW=10,nuW=11
```
usarica, line 479: What is the convention of ordering for lep/nu indices when either of the W charges are reversed?

usarica: (or if one were to take W->qq' decay)

MELA/fortran/mod_TTBHiggs.F90
```diff
@@ -6,6 +6,7 @@ MODULE ModTTBHiggs
 public :: EvalXSec_PP_TTBH, EvalXSec_PP_BBBH, EvalAmp_GG_TTBH, EvalAmp_QQB_TTBH
 public :: InitProcess_TTBH
 public :: ExitProcess_TTBH
+public :: ubarSpi_Dirac, vSpi_Dirac
```
usarica, line 9: Is this change necessary? If not, I would remove it.

usarica (changes requested):

MELA/fortran/mod_TTBHiggs.F90
tmartini: yes, it is necessary since the tHW Code uses the (massive) Dirac spinor subroutines from the ttbar code.

usarica: Ok, but in this case, mod_TTBHiggs needs to be compiled before tHW, which cannot be parallelized. I would expect 'make -j' to fail when this module is included unless the dependency is included explicitly (and to be frank, I don't like too many dependencies across modules that do completely different ME calculations; we tried to move away from it already). Is it possible to duplicate these functions?

MELA/fortran/mod_THiggs.F90
tmartini: due to charge conservation it is only possible to reverse both charges of both W bosons in t+H+W^- together resulting in the process tbar+H+W^+. The ordering is analogous to the assignment in ttbar (see mod_CrossSectionTH.F90 for particle IDs). 
In fact, all the routines in this function are valid for W->f'fbar (with f' and fbar being appropriate fermion/antifermion) and lep/nu are only placeholders.

MELA/fortran/mod_TTBHiggs.F90
tmartini: counter-proposal: I could move the spinor functions into mod_TopDecay.F90
I would not duplicate too many functions since I already had problems with duplicated functions when implementing the code in JHUGen.

MELA/fortran/mod_THiggs.F90
usarica: Sorry, then I don't understand the lep/nu vs sign convention. Which one of lep/nu indices are f and which ones are fbar? Are they reversed in t vs W decay by the opposite-charge pairing of Ws?

MELA/fortran/mod_TTBHiggs.F90
usarica: Possible, but please make sure makefile works then.

Would you please also run the full compilation chain and do

cd test
python testME_all.py

and let me know if there are errors?

MELA/fortran/mod_THiggs.F90
tmartini: for W^- the lep is the fermion and nu the antofermion.
for W^+ the opposite.

usarica: and the assumption in the ME coding is that it is a t(+) W(-) final state?

tmartini: depends on the ME.
There are two:
EvalAmp_GB_TWMH for t(+) W(-) H

and 

EvalAmp_GBB_TBWPH for tbar(-) W(+) H

cf. existing single top s-/t-channel

MELA/fortran/mod_TTBHiggs.F90
tmartini: > Ok, but in this case, mod_TTBHiggs needs to be compiled before tHW, which cannot be parallelized. I would expect 'make -j' to fail when this module is included unless the dependency is included explicitly (and to be frank, I don't like too many dependencies across modules that do completely different ME calculations; we tried to move away from it already). Is it possible to duplicate these functions?

since I have put "use ModTTBHiggs" in the ME routine which need the dirac spinor routines from ttbar, I would not expect the make -j to fail

usarica: Doesn't matter, the makefile doesn't know what use statements the code has. It only cares about the dependencies specified.

MELA/fortran/mod_THiggs.F90
usarica: Understood. Is it possible to have one or two lines of comments above these two functions to note the difference and the lep/nu convention as well so that when they are interfaced, the momenta passed are ordered correctly?

tmartini: I will do that. 
I will get back to it later in the (my) evening.
Thanks so far.

MELA/fortran/mod_TTBHiggs.F90
tmartini: Understood.
In the makefile in the Generator directory I have actually specified the dependencies accordingly. 
I just realized that I did not push the respective makefile to MELA yet. I will also do that later.

usarica: Good that we caught it then, thanks! Do you think you can also move the functions at the same time as you proposed?

MELA/fortran/mod_THiggs.F90
tmartini: I have inserted respective comments

MELA/fortran/mod_TTBHiggs.F90
tmartini: I have moved the dirac spinors to mod_TopDecay

tmartini: makefile doesn't need changing now

tmartini: > Possible, but please make sure makefile works then.
> 
> Would you please also run the full compilation chain and do
> 
> cd test
> python testME_all.py
> 
> and let me know if there are errors?

I cannot run the compilation chain for MELA here at home right now because I do not have ROOT installed on my laptop.
The exact same codes in the JHUGenerator directory did compile with the original makefile, though.

tmartini: I have tried to compile the MELA code on my office machine but I am still missing libraries:
libmcfm_707 and ljhugenmela
Can maybe one of you test the compilation chain?
Thanks

usarica (approved):
I tested the changes. It looks like they work, so I can approve.

gritsan (approved):

PR #9, opened 2021-07-05
Update reference files

usarica:
Include ZZ-Zgs mixture check for the ZZ final states in gluon fusion

### DIFF
--- a/MELA/BuildFile.xml
+++ b/MELA/BuildFile.xml
@@ -5,7 +5,7 @@
 <!--Use magic strings to avoid environment expansion in some shells. Replacements are done in TUtilHelpers::ExpandEnvironmentVariables. -->
 <flags CPPDEFINES="_melapkgpath_='.oODOLLAROo..oOOPEN_BRACKETOo.CMSSW_BASE.oOCLOSE_BRACKETOo./src/JHUGenMELA/MELA/'"/>
 <!--Some magic to retrieve the mcfm library over http, and link it hardcoding its path in the resulting library, so that LD_LIBRARY_PATH is not required. -->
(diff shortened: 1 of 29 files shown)


usarica:
This PR adds the ZZ-Zgs mixture checks on gg MCFM vs JHUGen.

Bottom line: There is agreement.

usarica:
Please approve so that we can merge.

usarica:
(Note: There are slightly different MCFM vs JHUGen ratios when flavor=4, which corresponds to Z1->qq summed, but that is because summation is performed without K factors in MCFM and with in JHUGen. This is known. Exclusive Z1->dd, Z->uu agrees using flavor=5.)

usarica:
Last commit is a CMS hotfix that should be approved.

gritsan (commented):
Ulascan: it is better to separate the PR with hot fixes from other things, otherwise this does not look clean and it is harder to track changes. However, to proceed faster without such separation, we need to discuss additional changes over Zoom or Skype. Among other things, we need to discuss if we need to propagate changes to the official release of the JHUGen package on the web page.  However, this discussion is not for Git.

usarica:
Sorry, but I have no reason to split this PR into two. The update to references were pending for the JHUGen release, and they do not change how MELA operates. I would suggest to approve this PR immediately. A discussion of changes to the JHUGen repository should happen there (and it is a trivial matter that does not even need a discussion tbh). In addition, matters related to policy should be outside of Git discussions, I agree, but matters related to code review should be on Git; Git is precisely designed to optimize it, and it is common for discussion to happen on this platform.

hroskes (approved):
I think this is fine, it's just additional tests and updates to the references.

PR #11, opened 2021-07-25
Update reference files

mostafa1993:
I also checked the compilation in `CMSSW_10_2_22` and it works fine.


usarica (approved):
The PR has the same commits as PR#9, but it is good that it is checked independently by a third person from CMS now. Given this, and the fact that review from hroskes is transferred, it can be merged.

PR #18, opened 2022-06-24
python3 MELA wrapper and new precompiled MCFM library 

jdavis36:
I made some changes to the data directory for easier installation of MELA on the CMSSW release we are currently using for analysis at JHU. I also include new scrips for calling the MELA python wrapper with python3 without overwriting the older scripts.

### DIFF
--- a/MELA/data/slc7_amd64_gcc900/download.url
+++ b/MELA/data/slc7_amd64_gcc900/download.url
@@ -1,0 +1,1 @@
+http://spin.pha.jhu.edu/Generator/MCFM-precompiled/v7_900/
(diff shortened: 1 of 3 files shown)


usarica (changes requested):
I would prefer Python wrapper update to be in a different commit. In general, I would like the python wrapper to be decomissioned.
Re. MCFM: does v7_920 not work?

jdavis36:
The current CMSSW release we are using is not available with gcc920 so to simplify the installation we just wanted to have a folder for gcc900 with a new download link. Did you mean we could use the mcfm library compiled for gcc920 as the download link for the new folder I added? Also in regards to the python wrapper being decommissioned, why do you want this to happen? I feel it is better to have that option for a potential user rather than not. once the python3 updates go in I do not foresee having to update the wrapper further unless something like python 4 gets released. I can still add the python changes in a separate pull request if you would prefer.

PR #19, opened 2022-06-27
Fix standalone compilation under gcc 10.x

usarica:
Opened per request of CMS colleagues, this PR fixes compilation issues with gcc versions 10.x (partially fixing compilation under CMSSW 12.x.y).

Only the standalone part is fixed as there are some ROOT dictionary compilation issues related to CMSSW 12.x.y. These issues do not appear in standalone compilation, so they are deemed as CMS software bugs at the moment. A later PR could be issued to fix specific issues if CMSSW-type compilation needs any fixes, but it is unnecessary at the moment.

### DIFF
--- a/MELA/.gitignore
+++ b/MELA/.gitignore
@@ -1,3 +1,4 @@
 obj
 lib
 executables
+**/TEMP
--- a/MELA/data/slc7_amd64_gcc10/download.url
+++ b/MELA/data/slc7_amd64_gcc10/download.url
@@ -1,0 +1,1 @@
+http://spin.pha.jhu.edu/Generator/MCFM-precompiled/v7_920/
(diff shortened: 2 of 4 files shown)


hroskes (approved):

gritsan:
I am traveling, so Jeff should be in a position to review and sign off if needed

gritsan:
Apparently neither Jeff nor I are allowed to review this code. Can this be fixed?

gritsan:
I confused code owner and review rights for myself, but Jeff needs to get review rights.

usarica:
> I confused code owner and review rights for myself, but Jeff needs to get review rights.

@gritsan: It appears as @jdavis36 already has maintenance rights. I did not see you, so I added you as well now.

usarica:
BTW, it looks like the most straightforward solution is to the CMSSW 12 problem is to remove CMSSW-specific computation. This doesn't mean anything critical will change on the user side; we can copy the MELA library to ${CMSSW_BASE}/lib ourselves (it already has the same name as what CMSSW would produce through scram). Same goes with the python directory - though I may eventually remove it later anyway.

If you agree to decommission CMSSW-specific compilation, I will update this PR. Otherwise, I will open a new one so that you can see the suggested change explicitly later.

PR #25, opened 2023-07-14
Fixed 13.6 -> 14 TeV rounding error

michael11892:
Fixed issue where using sqrts=13.6 would result in filenames with 14TeV, due to a rounding error in the way doubles were parsed into strings for the file path


usarica:
@michael11892 Thanks for pointing this out and proposing a solution. Once CMS provides 13.6 TeV high-mass samples for new calibration constants, we would be happy to fix this.

PR #26, opened 2023-07-14
Fixed 13.6 -> 14 TeV rounding error (interim)

michael11892:
Fixed issue where using sqrts=13.6 would result in filenames with 14TeV, due to a rounding error in the way doubles were parsed into strings for the file path

### DIFF
--- a/MELA/src/SuperDijetMela.cc
+++ b/MELA/src/SuperDijetMela.cc
@@ -35,6 +35,7 @@ void SuperDijetMela::Build(){
 void SuperDijetMela::SetupResolutionModel(TVar::Production prod){
   // Setup file path
   const string MELAPKGPATH = TVar::GetMELAPath();
+  double intpart;
   TString path = TString(MELAPKGPATH.c_str()) + "data/resolution_mJJ_recoVStrue_";
   TString prodName;
   switch (prod){
@@ -48,10 +49,12 @@ void SuperDijetMela::SetupResolutionModel(TVar::Production prod){
     MELAout << "SuperDijetMela::SetupResolutionModel: Production " << TVar::ProductionName(prod) << " is unknown." << endl;
     return;
   }
+  if (modf(sqrts, &intpart) == 0.0) floatflag = "%.0f"; else floatflag = "%.1f";
+  
   path += prodName;
-  path += Form("_%.0fTeV%s", sqrts, ".root");
+  path += Form("_"+floatflag+"TeV%s", sqrts, ".root");
 
-  TString appendName = Form("mJJReso_%.0fTeV", sqrts);
+  TString appendName = Form("mJJReso_"+floatflag+"TeV", sqrts);
   MELADifermionResolutionModel* model = new MELADifermionResolutionModel(prod, sqrts, path, appendName);
   if (model->isValid()){
     int iprod = (int)prod;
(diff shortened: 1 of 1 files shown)


usarica (changes requested):
At the moment, the calibration files for 13.6 (or 14) TeV do not exist in data/, so this PR would only make sense to merge once those calibration files are available. That depends on CMS production status, which has not even started yet.

gritsan (dismissed):
This change is really needed, thanks for making it!

gritsan:
> At the moment, the calibration files for 13.6 (or 14) TeV do not exist in data/, 
> so this PR would only make sense to merge once those calibration files are available. 
> That depends on CMS production status, which has not even started yet.

There are two issues: lack of 13.6 calibration and wrong treatment of energy. This PR fixes the code issue of integer energy expectation. We must approve this request, as it is needed for Run-3 studies urgently. The calibration files from 13.0 TeV could be used in the meantime.

gritsan:
This is a HOTFIX!!!

usarica:
> This PR fixes the code issue

No it doesn't as long as it refers to a file that includes 13.6TeV in the name, which does not exist yet. The proper interim fix would floor the float.

gritsan:
This is not a transparent solution, which confused every single person who looked at this.

usarica:
> This is not a transparent solution, which confused every single person who looked at this.

One could also consider scanning for the available sqrts values and then picking the closest one if the needed file is unavailable.. However, as the PR currently stands, it does not address the problem by opening the 13TeV files when sqrts=13.6 TeV. For that reason, the PR should be revised, whether the implemented solution is confusing or not.

hroskes (dismissed):

usarica (changes requested):
If you would like to pursue a non-searchable option, I think this optional argument is not needed. All you would need to do is something like.
int isqrts = sqrts;
...
Form("...%iTeV...", isqrts...)

Since this is an interim fix due to the way it is currently progressing, all tags need to be marked as 'pre-release' until a searchable solution is implemented, or we edit these lines again to accommodate 13p6TeV files that will be generated in the future, albeit at an undetermined time scale.

michael11892:
I see no merit in removing the proposed fix for the sake of a floor command which will be, by definition, overhauled at a later date. The additional option allows for the floor "interim fix" until production is ready.

usarica:
> I see no merit in removing the proposed fix for the sake of a floor command which will be, by definition, overhauled at a later date. The additional option allows for the floor "interim fix" until production is ready.

There is always a danger in putting code that is more complex than necessary. While you might not see a merit, the complexity of the code in this case has resulted in a bug already, which also tells me the commits are done without checking whether they work: L39 is set to false, and for the purpose of the interim fix for which it is implemented, it should have been set to true. That is why keeping the code as simple as possible helps.

Given that there are two solutions, one interim and one more permanent, I would suggest to split these into two separate PRs, with the permanent solution PR already starting from the interim solution PR so that both PRs can be merged in due time, and we won't have to remember or revisit these solutions in depth - and the fact that a PR is pending would act as a permanent reminder of its relevance when the CMS files are ready.

Once you split the PR, I would be happy to merge the interim version. Thanks.

michael11892:
The default value was intentionally set to false, that is no bug. However, your last proposed solution works nicely in this case. I will proceed to make one PR with the interim solution properly implemented and another for the interm fix's replacement, which can be merged in due time.

usarica (approved):

gritsan (approved):

PR #27, opened 2023-07-15
Fixed 13.6 -> 14 TeV rounding error (proper)

michael11892:
Replaced the previous hotfix (interim PR, #26). Sqrts=13.6 is properly parsed to path, rather than floored first. This PR is intended to be merged when sqrts=13.6 TeV production is ready.


usarica (changes requested):
Thanks, looks ok. I will block the merge until data files are ready.

PR #28, opened 2023-08-01
Added Anomalous Zff Couplings Interface to JHUGen & MCFM (C)

michael11892:
Added an interface that enables parsing anomalous Zff couplings to MCFM through MELA. The current PR only includes the necessary C infrastructure and does _not_ explicitly update the python Mela() object. All python updates relevant to the interface are scheduled for a separate PR.

### DIFF
--- a/MELA/interface/Mela.h
+++ b/MELA/interface/Mela.h
@@ -212,6 +212,7 @@ public:
     double selfDHvvcoupl_input[nSupportedHiggses][SIZE_HVV][2],
     double selfDHwwcoupl_input[nSupportedHiggses][SIZE_HVV][2],
     double selfDaTQGCcoupl_input[SIZE_ATQGC][2],
+    double selfDAZffcoupl_input[SIZE_AZff][2],
     float& prob,
     bool useConstant=true
     );
@@ -341,6 +342,8 @@ public:
   double selfDGvpvpcoupl[SIZE_GVV][2];
   //****aTQGC****//
   double selfDaTQGCcoupl[SIZE_ATQGC][2];
+  //****Anomnalous Zff**//
+  double selfDAZffcoupl[SIZE_AZff][2];
   // That is a lot of them!
 
   static void cleanLinkedFiles();
@@ -434,6 +437,7 @@ protected:
   void setSpinOneCouplings();
   void setSpinTwoCouplings();
   void setATQGCCouplings();
+  void setAZffCouplings();
 
   bool configureAnalyticalPDFs();
   void reset_SelfDCouplings();
--- a/MELA/interface/TCouplings.hh
+++ b/MELA/interface/TCouplings.hh
@@ -140,4 +140,19 @@ public:
   double aTQGCcoupl[SIZE_ATQGC][2];
 };
 
+class AZffCouplings{
+public:
+  AZffCouplings();
+  AZffCouplings(AZffCouplings const& other);
+  virtual ~AZffCouplings();
+
+  void reset();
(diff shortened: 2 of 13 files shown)


usarica (changes requested):
As I mentioned previously in the JHUGenerator commit, I do not think the anomalous Zff couplings approach is a good way to move forward. For that reason, I do not think it is a good idea to propagate it to MELA. Besides, the current architecture in MELA has been growing obsolete, so adding more 'functionalities' is not the best course of action at this moment. More discussion is necessary before we can move forward. If you need it for private studies, you can certainly use a private modification in your setup, but I would oppose making public releases.

michael11892 (commented):
Fixed merge conflict due to interim fix

gritsan:
We can have more discussion, but the outcome will not change: The generator options and the MELA option must match, because the users must be able to re-weight the generated samples, and to calculate discriminants. So, this discussion must focus on implementation, not on the idea of having this option, which has already been accepted to move forward. This is in reply to this comment:

> As I mentioned previously in the JHUGenerator commit, I do not think the anomalous Zff couplings approach is a good way to move forward. For that reason, I do not think it is a good idea to propagate it to MELA. Besides, the current architecture in MELA has been growing obsolete, so adding more 'functionalities' is not the best course of action at this moment. More discussion is necessary before we can move forward. If you need it for private studies, you can certainly use a private modification in your setup, but I would oppose making public releases.

usarica:
> We can have more discussion, but the outcome will not change: The generator options and the MELA option must match, because the users must be able to re-weight the generated samples, and to calculate discriminants. So, this discussion must focus on implementation, not on the idea of having this option, which has already been accepted to move forward. This is in reply to this comment:
> 
> > As I mentioned previously in the JHUGenerator commit, I do not think the anomalous Zff couplings approach is a good way to move forward. For that reason, I do not think it is a good idea to propagate it to MELA. Besides, the current architecture in MELA has been growing obsolete, so adding more 'functionalities' is not the best course of action at this moment. More discussion is necessary before we can move forward. If you need it for private studies, you can certainly use a private modification in your setup, but I would oppose making public releases.

I would suggest to move planning discussions offline. If the intention is for reweighting and discriminants to be computed, MELA is not the only place modifications should be made, and such discussion has larger scope and implications than what this PR attempts to address.

michael11892:
I have ran testME_all.py and testME_more.py with no errors returned for any test case. I have also tested the AZff couplings with the addition of ghzgs2=1 and checked the kinematic distributions to make sure that they matched the non-AZff results (to good approximation, the exact kinematic variables do have a subtle dependence on AZff) and the resulting distributions match our expectations, meaning that no bugs have been introduced into the code for independent ME calculations.

@usarica, you mentioned some cosmetic changes which you wanted to make. Given that the tests you provided are now concluded, it would be a good idea to tackle those changes before the final merge.

gritsan (approved):

gritsan (changes requested):
We really need to proceed with this interface, experts are waiting. Therefore:
Ulascan: please communicate your proposed cosmetic changes without so much delay. 
Michalis: please include updates to JHUGen matrix element interface to this PR, since it relies on the same changes as for MCFM. We cannot wait much longer for sequential updates.

gritsan (approved):

usarica (approved):
This is to close on this PR though #29 should be the one to approve.

michael11892:
Let's do things in an orderly fashion. Let's merge #28 and then #29.

gritsan (approved):

PR #33, opened 2023-09-25
updating to reflect JHUGen

MohitS704:
This is an update to reflect the changes for the random range factorization scheme in JHUGen in MELA


gritsan (dismissed):

usarica (changes requested):
There appears to be many lines changed in mod_Kinematics. This should be avoided.  Changes in scale schemes should be propagated to the C++ code as well, but this propagation needs to happen coherently in a random deterministic manner. I will also wait for the resolution of offline discussion pertaining to the physics use case and whether this is the desired implementation before merging. No merging can happen until that discussion is resolved.

gritsan:
All pending questions have been answered. 

> There appears to be many lines changed in mod_Kinematics. This should be avoided. Changes in scale schemes should be propagated to the C++ code as well, but this propagation needs to happen coherently in a random deterministic manner. I will also wait for the resolution of offline discussion pertaining to the physics use case and whether this is the desired implementation before merging. No merging can happen until that discussion is resolved.

MohitS704:
there - the changes are accurately reflected now

usarica (changes requested):
Please update the enums on https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TVar.hh#L195-L213.

MohitS704:
What do these enums do? I don't know what to put in there - I've never interacted with them before and they seem like an interface tool.

usarica:
> What do these enums do? I don't know what to put in there - I've never interacted with them before and they seem like an interface tool.

Yes, these enums are meant to provide the user the scale scheme choices in JHUGen. Even if the latest scheme is not applicable to MELA, the number of schemes should still be the same as what is in JHUGen. For this reason, I suggest you add the line,

`Dynamic_RandomUniform_Constrained,`

below

`Dynamic_Softest_pTJ,`

so that nEventScaleSchemes is incremented properly.

MohitS704:
The enums are updated

gritsan (approved):

usarica (approved):

PR #34, opened 2023-10-13
Creating the ability to change the Z Mass in JHUGenMELA

MohitS704:
Currently, the Z Mass can be changed in JHUGenerator but not in JHUGenMELA (without hardcoding). This push seeks to let you be able to do so via the function TUtil::SetJHUGenZMassWidth(MZ, GaZ), which is modelled after the near identical function TUtil::SetJHUGenHiggsMassWidth(MReso, GaReso).

The functionality from the SetMass and SetDecayWidth functions in the FORTRAN has not been changed - simply ported to the interface _should_ someone desire to change the Z Mass.

### DIFF
--- a/MELA/fortran/mod_JHUGenMELA.F90
+++ b/MELA/fortran/mod_JHUGenMELA.F90
@@ -11,6 +11,7 @@ MODULE ModJHUGenMELA
 
 public :: SetEWParameters
 public :: SetHiggsMassWidth
+public :: SetZMassWidth
 public :: SetDecayModes
 public :: SetTopDecays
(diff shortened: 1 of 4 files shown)


gritsan (approved):
This is needed for decay to charmonium final states

usarica:
This PR is not needed. Mela already has the functionality to change masses and widths of bosons and fermions in a way consistent between JHUGen and MCFM, see https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/Mela.h#L79-L80.

PR #40, opened 2023-11-28
Updating ResonancePropagatorScheme and Adding new gcc10 compiled MCFM library

lk11235:
Added the AltRunningWidth=4 to match previously accepted changes made to ResonancePropagatorScheme

### DIFF
--- a/MELA/data/slc7_amd64_gcc10/download.url
+++ b/MELA/data/slc7_amd64_gcc10/download.url
@@ -1,0 +1,1 @@
+http://spin.pha.jhu.edu/Generator/MCFM-precompiled/v7_10/
--- a/MELA/data/slc7_amd64_gcc10/download.url
+++ b/MELA/data/slc7_amd64_gcc10/download.url
@@ -1,1 +1,0 @@
-http://spin.pha.jhu.edu/Generator/MCFM-precompiled/v7_10/
--- a/MELA/interface/TVar.hh
+++ b/MELA/interface/TVar.hh
@@ -118,7 +118,8 @@ namespace TVar{
     NoPropagator=0,
     RunningWidth=1,
     FixedWidth=2,
-    CPS=3
+    CPS=3,
+    AltRunningWidth=4
   };
 
   enum Process{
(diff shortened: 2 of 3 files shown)


usarica (approved):

lk11235:
Pushed edits to the master branch without realizing that this PR was still open. But keeping these commits included here since they are also small necessary changes which were explicitly requested. 

Added a new download.url file and condition in the setup script to use the recently compiled and uploaded gcc10 MCFM library.

lk11235:
Sorry for the delay in commenting. JHUGen will compile and run with MELA using the gcc920 library.

usarica:
Great! So in that case, do we still need this PR, given that the MCFM library compiled under gcc 9.2.0 works in gcc 10?

jdavis36:
Hi All,

Just to add my two cents. I think the issue isn't so much that the we need an explicit gcc 10 version.  The issue is that if you run the setup script on gcc10, all of the MELA libraries would go into the gcc10 folder and the mcfm library goes into the gcc920 folder. So then when you export the library paths using 

`./setup env`

LD_LIBRARY_PATH does not have the path to the mcfm library. This is an easy fix for a developer, but for an end user its not immediately apparent that the mcfm library was dumped into a different directory. 

Could we change the steup.sh to put whatever mcfm library we use in the correct place, or make sure the environment setup script also includes the correct export path?

lk11235:
Yes, this is the obvious fix and one that has already been implemented in our privately compiled version of JHUGen for gridpack production. I will update this PR with the relevant changes and remove the unnecessary gcc10 library once I'm back on a computer.

usarica:
@jdavis36 @lk11235 I think the best solution is to remove these lines altogether: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/setup.sh#L56-L58. That way, SCRAM_ARCH will never be referenced anywhere in the JHUGen framework. (Do we accidentally use this environment variable in the Lexicon?)

gritsan:
Please close on this PR, we need to move forward

usarica:
> Please close on this PR, we need to move forward

Waiting for @lk11235 to take the requested action above.

lk11235:
Making the aforementioned edits and will be testing the full JHUGen compilation from scratch

lk11235:
MELA compilation has been tested and works with the requested changes. This PR should be good to merge now

gritsan (approved):

gritsan:
We need to release a new version of the generator ASAP before the holidays. This is needed by the BPH group on CMS, with changes introduced by Mohit. Could everybody please make an effort to close the loop?

usarica:
@gritsan This PR is fine, but there are still things to fix in Mohit's PR.

usarica (approved):

PR #42, opened 2024-01-09
Updating mela.py to Python 3

MohitS704:
Given the deprecated and unsupported status of Python 2, this fix moves `mela.py` and `pythonmelautils.py` from Python 2 to Python 3. 

Similarly, values for the 4th generation t and b quark couplings were added into `mela.py`, and the whitespace was changed to match up to the industry standard of tabs that are 4 spaces in length.


gritsan (approved):

MohitS704:
@gritsan @usarica I believe that this should warrant a new version of MELA - given we are upgrading to a whole new version of a language.

MohitS704:
Once we agree on that I'll merge the PR

PR #44, opened 2024-02-13
Adding RenFacScheme Getter

MohitS704:
There is currently no way to access the current factorization scheme that MELA uses at any given point. This fixes that.

I've mimicked the pipeline of set_RenFacScheme to have get_RenFacScheme (it goes through ZZMatrixElement then through TUtil), where it returns the TVar::event_scales_type that someone can then access on their own.


gritsan (dismissed):

MELA/src/Mela.cc
```diff
@@ -478,6 +478,9 @@ void Mela::setRemoveJetMasses(bool MasslessLeptonSwitch){ TUtil::applyJetMassCor
 void Mela::setRenFacScaleMode(TVar::EventScaleScheme renormalizationSch, TVar::EventScaleScheme factorizationSch, double ren_sf, double fac_sf){
   ZZME->set_RenFacScaleMode(renormalizationSch, factorizationSch, ren_sf, fac_sf);
 }
+TVar::event_scales_type Mela::get_RenFacScaleMode(){
```
usarica, line 481: get_ -> get

MELA/src/TEvtProb.cc
```diff
@@ -207,11 +207,18 @@ void TEvtProb::SetVerbosity(TVar::VerbosityLevel tmp){ verbosity = tmp; }
 void TEvtProb::SetLeptonInterf(TVar::LeptonInterference tmp){ leptonInterf = tmp; }
 void TEvtProb::SetCandidateDecayMode(TVar::CandidateDecayMode mode){ PDGHelpers::setCandidateDecayMode(mode); }
 void TEvtProb::SetRenFacScaleMode(TVar::EventScaleScheme renormalizationSch, TVar::EventScaleScheme factorizationSch, double ren_sf, double fac_sf){
+  if(verbosity>=TVar::DEBUG_VERBOSE){
+    MELAout << "Renormalization Scheme/Scale Factor set to " << renormalizationSch << "/" << ren_sf << endl;
+    MELAout << "Factorization Scheme/Scale Factor set to " << factorizationSch << "/" << fac_sf << endl;
+  }
```
usarica, line 213: I would either remove these or remake the reference files in case any of them call SetRenFacScaleMode.

usarica (changes requested):

usarica (changes requested):
One more request: get functions should have the const keyword associated. If possible, they should also return const-reference (I leave it to you to find out why).

MohitS704:
> One more request: get functions should have the const keyword associated. If possible, they should also return const-reference (I leave it to you to find out why).

This has been done.

MELA/interface/Mela.h
```diff
@@ -96,6 +96,7 @@ class Mela{
   void getConstant(float& prob); // <ME> constants
   void getPAux(float& prob); // SuperProb
 
+  const TVar::event_scales_type& getRenFacScaleMode();
```
usarica, line 99: Should be

`const TVar::event_scales_type& getRenFacScaleMode() **const**;`

(Similar everywhere else)

usarica (changes requested):

usarica:
> > One more request: get functions should have the const keyword associated. If possible, they should also return const-reference (I leave it to you to find out why).
> 
> This has been done.

What I meant was the function qualifier, not just the qualifier of its return type.

usarica (approved):

gritsan (approved):

usarica (changes requested):
Function signatures also need to match in source files.

usarica:
Please reopen a new PR reverting #46 and implementing the above comment.

In general, please do not push and/or flag as done before you are sure MELA actually compiles.

PR #45, opened 2024-02-28
Making a direct binding for python

MohitS704:
This change makes it so that using python for MELA is done via a direct link to the methods and fields that exist within the C++, rather than loading it via ROOT. The makefile generates a CPython shared library via pybind11 (which is included with CMSSW). This change makes the python more robust and cleaner, while at the same time keeping the exact same methods and values as the C++.


gritsan (approved):

hroskes (dismissed):
Thanks a lot Mohit for cleaning up my messy code!

MELA/python/mela_binding.cpp
```diff
@@ -0,0 +1,5145 @@
...
+#include "TMCFM.hh"
+#include "TLorentzVector.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+#include "pybind11/numpy.h"
+namespace py = pybind11;
+using namespace std;
+
+namespace pymela{
+    enum HQQ_indices{
```
usarica, line 13: Until L215 and for later in other index duplications:

I think it might be better to simply name the enums in TCouplingBase.hh as
CouplingIndex_*, where * is consistent with SIZE_*, e.g.

```
enum CouplingIndex_HQQ{
    gHIGGS_KAPPA,
    gHIGGS_KAPPA_TILDE,

    SIZE_HQQ
};
```

You can then use

'using HQQ_indices = ::CouplingIndex_HQQ',

which would avoid duplicating the indices unnecessarily.

Now, this would not avoid the duplications downstream (value(...) statements). In order to do that, I think it might be better to convert each index set into a set of macros, e.g.

```
#define MELACMDS_INDEX_HQQ \
  MELACMD_INDEX_HQQ(gHIGGS_KAPPA) \
  MELACMD_INDEX_HQQ(gHIGGS_KAPPA_TILDE)
```

which allows you to write later something like

```
#define MELACMD_INDEX_HQQ(COUPLING) COUPLING,
enum CouplingIndex_HQQ{ MELACMDS_INDEX_HQQ SIZE_HQQ };
#undef MELACMD_INDEX_HQQ
...
#define MELACMD_INDEX_HQQ(COUPLING) .value(#COUPLING, pymela::COUPLING)
    py::enum_<pymela::HQQ_indices>(m, "HQQ_indices") MELACMDS_INDEX_HQQ .value("SIZE_HQQ", pymela::SIZE_HQQ);
#undef MELACMD_INDEX_HQQ
```

MELA/python/mela_binding.cpp
```diff
@@ -0,0 +1,5145 @@
...
+        .value("gAZff_ZnunuRH", pymela::gAZff_ZnunuRH)
+        .value("gAZff_ZnunuLH", pymela::gAZff_ZnunuLH)
+        .value("gAZff_uZRH", pymela::gAZff_uZRH)
+        .value("gAZff_uZLH", pymela::gAZff_uZLH)
+        .value("gAZff_dZRH", pymela::gAZff_dZRH)
+        .value("gAZff_dZLH", pymela::gAZff_dZLH)
+        .value("SIZE_AZff", pymela::SIZE_AZff);
+
+
+}
```
usarica, line 5145: Missing end-of-line.

usarica (changes requested):
In general, I do not like the amount of code duplicated. It makes it difficult to synchronize quickly, and we did a huge amount of work to avoid duplicates during the development of Mela v2.

I have provided a review with example implementations that you could follow to clean up the code.

We cannot have 5145-line file and expect future developers to implement changes, or anyone to realistically read it. The fact that it is full of duplications means it can be enormously simplified. You should be able to reduce the visible size down to the level of <1000 lines (perhaps only a few hundred if you try hard enough and modify my suggestion slightly).

MELA/makefile
```diff
@@ -28,6 +28,10 @@ MELAPKGPATH = $(MELADIR)/
 LIB = libJHUGenMELAMELA.so
 LIBRULE = $(MELALIBDIR)/$(LIB)
 
+pyLIB = Mela$(shell python3-config --extension-suffix)
```
usarica, line 31: python-config/python3-config requires python-dev package. You might want to implement an alternative name specification.

MELA/makefile
```diff
@@ -82,6 +86,12 @@ $(LIBRULE):	$(OBJECTS) $(MELAOBJDIR)/LinkDef_out.o | alldirs
 	echo "Linking $(LIB)"; \
 	$(LINKER) $(LINKERFLAGS) -shared $(OBJECTS) $(MELAOBJDIR)/LinkDef_out.o -o $@
 
+$(pyLIBRULE): $(pyLINKFILE) | alldirs
+	@echo "Making python"; \
+	$(LINKER) $(pyLINKFILE) $(CPPFLAGS) $(LINKERFLAGS) \
+	-O3 -Wall -shared $(OBJECTS) $(MELAOBJDIR)/LinkDef_out.o -std=c++17 -fPIC \
+	$(shell python3 -m pybind11 --includes) -o $@
```
usarica, line 93: Also over here, conditionally compile when python 3.x is detected (I assume the bindings are python 3 specific).

MELA/python/mela_binding.cpp
MohitS704: This solution is quite welcome! I was looking around for something to prevent duplicating the code (I realized partway it was incredibly messy but  couldn't think of a clean solution) - so thank you. I'll do my best to implement this and clean up what I can.

MohitS704: One question here - 

```#define MELACMDS_INDEX_HQQ \
  MELACMD_INDEX_HQQ(gHIGGS_KAPPA) \
  MELACMD_INDEX_HQQ(gHIGGS_KAPPA_TILDE)```

Is still duplicating code, albeit in a different way. One would still need to keep track of any/all couplings that could be added in the future. The fact that they are now defined as macros instead of explicitly as enumeration value arguments is now another layer of abstraction - I'm not sure this is the way to go.

I **will** look at using the macros to define the plethora of MELA couplings, however. That certainly seems useful.

Comment (usarica) [reply MELA/python/mela_binding.cpp:13]: Well, you need to define the coupling somewhere.
```
#define MELACMDS_INDEX_HQQ \
  MELACMD_INDEX_HQQ(kappa, gHIGGS_KAPPA) \
  MELACMD_INDEX_HQQ(kappa_tilde, gHIGGS_KAPPA_TILDE)
  
#define MELACMD_INDEX_HQQ(COUPLINGSTR, COUPLING) COUPLING,
enum CouplingIndex_HQQ{ MELACMDS_INDEX_HQQ SIZE_HQQ };
#undef MELACMD_INDEX_HQQ
```
should be sufficient to declare the enums once. For the rest, you can use `COUPLINGSTR` or `COUPLING`, or both, within the `#define`-`#undef` blocks of `MELACMD_INDEX_HQQ`.

MELA/makefile
MohitS704: python-dev is a dependency of pybind11 [see here](https://pybind11.readthedocs.io/en/stable/basics.html#linux-macos)

If the person doesn't have pybind11 installed then this is a moot point. I can change the python check in setup.sh to instead look for an installation of pybind11

MELA/python/mela_binding.cpp
MohitS704: I've simply kept the enums as is, but now pull directly from the named TCouplingsBase indices

MohitS704:
All couplings have been updated to use C++ macros instead of repetitive code.

As far as the enums go - we are required to list them once, and since we now pull directly from named enums in `TCouplingsBasehh`, it makes the most sense for someone to add to the enums directly rather than interact with another layer of abstraction.

MELA/makefile
MohitS704: This check now happens in setup.sh - resolved.

hroskes (dismissed):

gritsan (approved):

MohitS704:
@usarica I've seen no further comments from you in the last 2 days since I made the edits that you requested to shorten the length of the file. I've replied to your concerns with my own reasons, and so far I see no issues. Can we please merge this?

gritsan:
We have two approvals and pending requests addressed, the block for merging needs to be lifted, if there are no further comments

usarica:
I was not sure if the changes were complete; not all of my suggestions seem to have been implemented. Could @MohitS704 elaborate?

MohitS704:
To your comment about the enums here:
>Well, you need to define the coupling somewhere.
>
>#define MELACMDS_INDEX_HQQ \
>  MELACMD_INDEX_HQQ(kappa, gHIGGS_KAPPA) \
>  MELACMD_INDEX_HQQ(kappa_tilde, gHIGGS_KAPPA_TILDE)
>  
>#define MELACMD_INDEX_HQQ(COUPLINGSTR, COUPLING) COUPLING,
>enum CouplingIndex_HQQ{ MELACMDS_INDEX_HQQ SIZE_HQQ };
>#undef MELACMD_INDEX_HQQ
>
>should be sufficient to declare the enums once. For the rest, you can use COUPLINGSTR or COUPLING, or both, within the #define-#undef blocks of MELACMD_INDEX_HQQ.

I replied as such:

>As far as the enums go - we are required to list them once, and since we now pull directly from named enums in TCouplingsBasehh, it makes the most sense for someone to add to the enums directly rather than interact with another layer of abstraction.

I can elaborate. You correctly state that we need to define the enums somewhere - and you also state that this is for the sanity of the developer. We have named the enums in TCouplingsBase according to your naming convention, and it is my opinion, as the original developer of this code, that keeping the explicit enum values within the enumeration for the python is the best way to go about this since it is the simplest. Wrapping up a list in a preprocessor function is another unnecessary layer of abstraction.

MohitS704:
To your comment about the end of the line:

>Missing end-of-line.

This is fixed in the `mela_binding.cpp` file

MohitS704:
To your comment about dependences, here:

>python-config/python3-config requires python-dev package. You might want to implement an alternative name specification.

I replied:

>python-dev is a dependency of pybind11 [see here](https://pybind11.readthedocs.io/en/stable/basics.html#linux-macos)
>
>If the person doesn't have pybind11 installed then this is a moot point. I can change the python check in setup.sh to instead look for an installation of pybind11

Let me elaborate on this point as well - if someone doesn't have pybind11 installed, they also don't have python-dev installed. If they don't have pybind11, then the binding doesn't work to begin with, and that entire check is useless. See my next comment where I introduce a new check. _All of these packages exist within CMSSW_

MohitS704:
To your followup here:

>Also over here, conditionally compile when python 3.x is detected (I assume the bindings are python 3 specific).

I replied:

> This check now happens in setup.sh - resolved.

To elaborate, I now check in setup.sh if you have the requisite packages installed - namely pybind11 and python3 - by calling python and importing pybind11 through the command line (silently), then checking the exit status of the command. You can see for yourself in `setup.sh` if you would like - it is the function on line 26 and called in line 66.

MohitS704:
Now to tackle the heart of the issue:

>In general, I do not like the amount of code duplicated. It makes it difficult to synchronize quickly, and we did a huge amount of work to avoid duplicates during the development of Mela v2.
>
>I have provided a review with example implementations that you could follow to clean up the code.
>
>We cannot have 5145-line file and expect future developers to implement changes, or anyone to realistically read it. The fact that it is full of duplications means it can be enormously simplified. You should be able to reduce the visible size down to the level of <1000 lines (perhaps only a few hundred if you try hard enough and modify my suggestion slightly).


I replied:

>All couplings have been updated to use C style macros instead of repetitive code.

The vast majority of the duplications occurred when couplings were defined (after all, we _do_ want this code to be backwards compatible with the old version of pyMELA) - these have been fixed by moving the code to the preprocessor, something inspired by your suggestion for the enums. Now, instead of the couplings being about 4100 lines, they are < 1000, exactly as you describe. I have already explained why I believe the enums should remain the way they are in a previous comment, so that part can be ignored.

MohitS704:
Now that the file size has been cut down by nearly 4000 lines, the code is much less confusing to a developer, and there are now checks in place within `setup.sh` to ensure that every dependency exists. I believe that solves all of the issues you brought up @usarica.

MohitS704:
Please also keep in mind that from line 410 to line 1112, nearly every line is accompanied by an empty space to help demarcate couplings to the reading developer - this is a stylistic choice to help maintain the sanity of someone reading the code. However, it means that the actual number of lines of code is almost halved in this region.

usarica:
> To your comment about the enums here:
> 
> > Well, you need to define the coupling somewhere.
> > #define MELACMDS_INDEX_HQQ 
> > MELACMD_INDEX_HQQ(kappa, gHIGGS_KAPPA) 
> > MELACMD_INDEX_HQQ(kappa_tilde, gHIGGS_KAPPA_TILDE)
> > #define MELACMD_INDEX_HQQ(COUPLINGSTR, COUPLING) COUPLING,
> > enum CouplingIndex_HQQ{ MELACMDS_INDEX_HQQ SIZE_HQQ };
> > #undef MELACMD_INDEX_HQQ
> > should be sufficient to declare the enums once. For the rest, you can use COUPLINGSTR or COUPLING, or both, within the #define-#undef blocks of MELACMD_INDEX_HQQ.
> 
> I replied as such:
> 
> > As far as the enums go - we are required to list them once, and since we now pull directly from named enums in TCouplingsBasehh, it makes the most sense for someone to add to the enums directly rather than interact with another layer of abstraction.
> 
> I can elaborate. You correctly state that we need to define the enums somewhere - and you also state that this is for the sanity of the developer. We have named the enums in TCouplingsBase according to your naming convention, and it is my opinion, as the original developer of this code, that keeping the explicit enum values within the enumeration for the python is the best way to go about this since it is the simplest. Wrapping up a list in a preprocessor function is another unnecessary layer of abstraction.

Thanks for the replies. In general, I do not find the mindset of the 'original developer' productive or welcoming; this is not software stack development review, and all of us contribute at different levels. If I were to keep this mindset, I could produce the same arguments on my end since I am reviewing your modification on the code that I wrote during the large scale revision of Mela for v2, and we wouldn't be able to move anywhere.

I am not sure why you argue it is an unnecessary layer of abstraction; what you find as unnecessary might not hold for others - at the level of the changes you have already provided, going further is a matter of taste, and you can simply state that you think the implementation is adequate enough to not warrant further code length reduction.

usarica:
> To your comment about dependences, here:
> 
> > python-config/python3-config requires python-dev package. You might want to implement an alternative name specification.
> 
> I replied:
> 
> > python-dev is a dependency of pybind11 [see here](https://pybind11.readthedocs.io/en/stable/basics.html#linux-macos)
> > If the person doesn't have pybind11 installed then this is a moot point. I can change the python check in setup.sh to instead look for an installation of pybind11
> 
> Let me elaborate on this point as well - if someone doesn't have pybind11 installed, they also don't have python-dev installed. If they don't have pybind11, then the binding doesn't work to begin with, and that entire check is useless. See my next comment where I introduce a new check. _All of these packages exist within CMSSW_

Just one note here: While in this case it does not matter much, CMSSW is *not* the only way to use MELA, so while checking everything through CMS software collections is useful, it is nevertheless a limited check in terms of whom else (ATLAS? Theory colleagues?) could use the tools.

usarica:
> Now to tackle the heart of the issue:
> 
> > In general, I do not like the amount of code duplicated. It makes it difficult to synchronize quickly, and we did a huge amount of work to avoid duplicates during the development of Mela v2.
> > I have provided a review with example implementations that you could follow to clean up the code.
> > We cannot have 5145-line file and expect future developers to implement changes, or anyone to realistically read it. The fact that it is full of duplications means it can be enormously simplified. You should be able to reduce the visible size down to the level of <1000 lines (perhaps only a few hundred if you try hard enough and modify my suggestion slightly).
> 
> I replied:
> 
> > All couplings have been updated to use C style macros instead of repetitive code.
> 
> The vast majority of the duplications occurred when couplings were defined (after all, we _do_ want this code to be backwards compatible with the old version of pyMELA) - these have been fixed by moving the code to the preprocessor, something inspired by your suggestion for the enums. Now, instead of the couplings being about 4100 lines, they are < 1000, exactly as you describe. I have already explained why I believe the enums should remain the way they are in a previous comment, so that part can be ignored.

Defining couplings by preprocessor arguments should not have a visible effect for the end-user. It is good that we are now within 1000 lines of code - it was not a hard limit, but it is good to see this reduction in visible code size (note 'visible' and not actual, 'compiled' code).

usarica (changes requested):
Python test scripts in the test directory also need to be checked and adjusted as appropriately. In particular, testME_more.py seems to refer to the mela file that no longer exists.

MohitS704:
> > To your comment about dependences, here:
> > > python-config/python3-config requires python-dev package. You might want to implement an alternative name specification.
> > 
> > 
> > I replied:
> > > python-dev is a dependency of pybind11 [see here](https://pybind11.readthedocs.io/en/stable/basics.html#linux-macos)
> > > If the person doesn't have pybind11 installed then this is a moot point. I can change the python check in setup.sh to instead look for an installation of pybind11
> > 
> > 
> > Let me elaborate on this point as well - if someone doesn't have pybind11 installed, they also don't have python-dev installed. If they don't have pybind11, then the binding doesn't work to begin with, and that entire check is useless. See my next comment where I introduce a new check. _All of these packages exist within CMSSW_
> 
> Just one note here: While in this case it does not matter much, CMSSW is _not_ the only way to use MELA, so while checking everything through CMS software collections is useful, it is nevertheless a limited check in terms of whom else (ATLAS? Theory colleagues?) could use the tools.

This is a fair point - I was just trying to point out that it isn't that large of a package to consider - nothing else

MohitS704:
> > To your comment about the enums here:
> > > Well, you need to define the coupling somewhere.
> > > #define MELACMDS_INDEX_HQQ
> > > MELACMD_INDEX_HQQ(kappa, gHIGGS_KAPPA)
> > > MELACMD_INDEX_HQQ(kappa_tilde, gHIGGS_KAPPA_TILDE)
> > > #define MELACMD_INDEX_HQQ(COUPLINGSTR, COUPLING) COUPLING,
> > > enum CouplingIndex_HQQ{ MELACMDS_INDEX_HQQ SIZE_HQQ };
> > > #undef MELACMD_INDEX_HQQ
> > > should be sufficient to declare the enums once. For the rest, you can use COUPLINGSTR or COUPLING, or both, within the #define-#undef blocks of MELACMD_INDEX_HQQ.
> > 
> > 
> > I replied as such:
> > > As far as the enums go - we are required to list them once, and since we now pull directly from named enums in TCouplingsBasehh, it makes the most sense for someone to add to the enums directly rather than interact with another layer of abstraction.
> > 
> > 
> > I can elaborate. You correctly state that we need to define the enums somewhere - and you also state that this is for the sanity of the developer. We have named the enums in TCouplingsBase according to your naming convention, and it is my opinion, as the original developer of this code, that keeping the explicit enum values within the enumeration for the python is the best way to go about this since it is the simplest. Wrapping up a list in a preprocessor function is another unnecessary layer of abstraction.
> 
> Thanks for the replies. In general, I do not find the mindset of the 'original developer' productive or welcoming; this is not software stack development review, and all of us contribute at different levels. If I were to keep this mindset, I could produce the same arguments on my end since I am reviewing your modification on the code that I wrote during the large scale revision of Mela for v2, and we wouldn't be able to move anywhere.
> 
> I am not sure why you argue it is an unnecessary layer of abstraction; what you find as unnecessary might not hold for others - at the level of the changes you have already provided, going further is a matter of taste, and you can simply state that you think the implementation is adequate enough to not warrant further code length reduction.

Apologies for that - you make a good point here. I was trying to emphasize an argument about the code reduction here - no offense intended.

MohitS704:
> Python test scripts in the test directory also need to be checked and adjusted as appropriately. In particular, testME_more.py seems to refer to the mela file that no longer exists.

I totally forgot about this - yes they do need to be checked. I will update those accordingly.

MohitS704:
> Python test scripts in the test directory also need to be checked and adjusted as appropriately. In particular, testME_more.py seems to refer to the mela file that no longer exists.

I've made a series of edits to testME_more.py. Nothing else seems to refer to mela.py within the tests. I think with this, can we close on this PR @usarica? 

Apologies for the general pushiness I've exhibited, but I'm trying to get this out before the workshop (pyMELA will be shown as an option regardless of whether this goes out in time, but this is certainly a much cleaner version that people can work with more easily in a tutorial).

hroskes (approved):

usarica:
@MohitS704 Thanks for the changes.

I am not sure whether people who are already using MelaAnalytics would adopt the python interface (some I know have already developed a python interface to this package), but you could certainly include this interface during the workshop.

With that, I think the PR is ok to merge.

usarica (approved):

PR #51, opened 2024-04-17
Grab Bag of Fixes

### LINKED ISSUE #49: Debugging features that need to be implemented in the MELA Python wrapper
Hi @MohitS704,

As mentioned during the LPC workshop tutorial, the Python wrapper needs to include debugging features. Specifically, we need to be able to call https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TUtil.hh#L332-L333.

Would you mind submitting a PR to add a wrapper for these calls?

Thanks!

MohitS704:
There are a few portions to this PR:

## The following changes have been made to the Python binding
- As stated by @usarica during the offshell workshop, the 0 has a special place in `Mela::setInputEvent`. As such, the function has been turned into a wrapper which then provides a 0 to the input function if given an empty simpleParticleCollection
- The function TUtil::PrintCandidateSummary (whose implentation closes #49) , which has 2 possible inputs, has been split into 2
  - PrintCandidateSummary is now the one that takes in a simple event record
  - PrintCurrentCandidateSummary takes the current MELA candidate and prints it out
- SimpleParticle_t has some QoL edits
  - Since TLorentzVectors are near impossible to convert using Pybind (PyROOT TLorentzVectors cannot be passed in as the same object in pybind11 due to PyROOT's implementation), methods for returning the vector of a SimpleParticle_t in <Pt, Eta, Phi, M> and <Px, Py, Pz, E> as tuples have been implemented
  - A `__repr__` function was added for easier debugging
- SimpleParticleCollection_t has pickling support included. This is to allow for data processing before any MELA call and for usage within the [MultiProcessing](https://docs.python.org/3/library/multiprocessing.html) package
- cleanLinkedFiles as well as calculate4Momentum have been added as functions that can be called from the Python
- Every Enum in TVar.hh is now reflected in the Python

## The following changes have been made to TVar.hh
- The previously unnamed enum was named to KAssociated to allow for binding to the Python

## The following changes have been made to setup.sh and the data folder
- el9 gcc12 support has been added (I have found out that this was the cause of the errors from the workshop - the gcc version for CMSSW 14 on el9 was missing for MCFM)
- See the updated download.url in the folder

### DIFF
--- a/MELA/interface/TVar.hh
+++ b/MELA/interface/TVar.hh
@@ -26,7 +26,7 @@ typedef std::vector<SimpleParticle_t> SimpleParticleCollection_t;
 
 
 namespace TVar{
-  enum{
+  enum KAssociated{
     kNoAssociated=1,
     kUseAssociated_Leptons=2, // l or nu
     kUseAssociated_Photons=3,
--- a/MELA/interface/TVar.hh
+++ b/MELA/interface/TVar.hh
@@ -26,7 +26,7 @@ typedef std::vector<SimpleParticle_t> SimpleParticleCollection_t;
 
 
 namespace TVar{
-  enum KAssociated{
+  enum{
     kNoAssociated=1,
     kUseAssociated_Leptons=2, // l or nu
     kUseAssociated_Photons=3,
--- a/MELA/python/mela_binding.cpp
+++ b/MELA/python/mela_binding.cpp
@@ -2,6 +2,8 @@
 #include "TVar.hh"
 #include "TCouplingsBase.hh"
 #include "TMCFM.hh"
+#include "TUtil.hh"
+#include "MELACandidate.h"
 #include "TLorentzVector.h"
 #include "pybind11/pybind11.h"
 #include "pybind11/stl.h"
@@ -201,7 +203,7 @@ SimpleParticleCollection_t collection_initializer_from_column(std::vector<int> i
             particle_initializer(ids[i], x[i], y[i], z[i], e[i], ptEtaPhi)
         );
     }
-    
+
     return collection;
 }
 
@@ -213,6 +215,29 @@ SimpleParticleCollection_t collection_initializer(py::list listOfParticles){
         collection.push_back(particle);
     }
     return collection;
+}
+
+void setInputEvent(Mela& mela, SimpleParticleCollection_t* daughters, SimpleParticleCollection_t* associated, SimpleParticleCollection_t* mothers, bool isgen){
+    if(daughters->size() == 0){
+        daughters = 0;
+    }
+    if (associated->size() == 0){
+        associated = 0;
+    }
+    if (mothers->size() == 0){
+        mothers = 0;
+    }
+    mela.setInputEvent(daughters, associated, mothers, isgen);
+}
+
+void PrintCurrentCandidateSummary(Mela& mela){
+    MELACandidate* curCand = mela.getCurrentCandidate();
+    TUtil::PrintCandidateSummary(curCand);
+}
+
+void PrintCandidateSummary(Mela& mela, TVar::simple_event_record curCand){
+    TVar::simple_event_record* curRecord = &curCand;
+    TUtil::PrintCandidateSummary(curRecord);
 }
 
 #define MAKE_COUPLING_ARR_SPIN_ZERO(arrayName, size, arrType)\
@@ -299,12 +324,14 @@ PYBIND11_MODULE(Mela, m) {
         .def_property_readonly("id", [](SimpleParticle_t& P){
             return P.first;
         })
-        .def_property_readonly("vector", [](SimpleParticle_t& P){
-            P.second.Print();
+        .def_property_readonly("PxPyPzE_vector", [](SimpleParticle_t& P){
             return py::make_tuple(P.second.Px(), P.second.Py(), P.second.Pz(), P.second.E());
         })
+        .def_property_readonly("PtEtaPhiM_vector", [](SimpleParticle_t& P){
+            return py::make_tuple(P.second.Pt(), P.second.Eta(), P.second.Phi(), P.second.M());
+        })
         .def("__repr__",[](SimpleParticle_t& P){
-            return "Particle with id " + std::to_string(P.first);
+            return std::to_string(P.first) + ": " + std::to_string(P.second.Px()) + ", " + std::to_string(P.second.Py()) + ", " + std::to_string(P.second.Pz()) + ", " + std::to_string(P.second.E());
         });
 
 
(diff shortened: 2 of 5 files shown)


MELA/setup.sh
```diff
@@ -16,6 +16,8 @@ getMELAARCH(){
     echo slc7_amd64_gcc820
   elif [[ "$GCCVERSION" == "8"* ]]; then
     echo slc7_amd64_gcc830
+  elif [[ "$GCCVERSION" == "12"* ]]; then
```
usarica, line 19: Do we really need this?

MELA/python/mela_binding.cpp
```diff
@@ -215,6 +217,29 @@ SimpleParticleCollection_t collection_initializer(py::list listOfParticles){
...
+    }
+    mela.setInputEvent(daughters, associated, mothers, isgen);
+}
+
+void PrintCurrentCandidateSummary(Mela& mela){
+    MELACandidate* curCand = mela.getCurrentCandidate();
+    TUtil::PrintCandidateSummary(curCand);
+}
+
+void PrintCandidateSummary(Mela& mela, TVar::simple_event_record curCand){
```
usarica, line 238: I don't think a Mela object needs to be passed here.

hroskes (approved):

MELA/python/mela_binding.cpp
```diff
@@ -215,6 +217,29 @@ SimpleParticleCollection_t collection_initializer(py::list listOfParticles){
...
+    if ((associated) && (associated->size() == 0)){
+        associated = 0;
+    }
+    if ((mothers) && (mothers->size() == 0)){
+        mothers = 0;
+    }
+    mela.setInputEvent(daughters, associated, mothers, isgen);
+}
+
+void PrintCurrentCandidateSummary(Mela& mela){
```
usarica, line 233: Ok, but we should still have a PrintCandidateSummary version which receives a MELACandidate. MELACandidates do not need to be created within Mela (and for complex applications, they are not).

MELA/python/mela_binding.cpp
usarica: Pertaining to the comment on L233, you could make another version that receives MELACandidate. You would be overloading the function, so make sure the type is disambiguated correctly.

MELA/python/mela_binding.cpp
```diff
@@ -215,6 +217,29 @@ SimpleParticleCollection_t collection_initializer(py::list listOfParticles){
...
+void setInputEvent(Mela& mela, SimpleParticleCollection_t* daughters, SimpleParticleCollection_t* associated, SimpleParticleCollection_t* mothers, bool isgen){
+    if((daughters) && (daughters->size() == 0)){
+        daughters = 0;
+    }
+    if ((associated) && (associated->size() == 0)){
+        associated = 0;
+    }
+    if ((mothers) && (mothers->size() == 0)){
+        mothers = 0;
+    }
```
usarica, line 229: Style comments:
- `(daughters) &&` is the same thing as `daughters &&`; no need for parentheses. Same for the `size() == 0` checks.
- Try to use `nullptr` instead of `0` for pointers. They end up being the same thing, but it allows better interpretability.
- Instead of `size() == 0`, it is preferable to use `empty()`.

Question: Does binding work if I pass `None` for, say, `mothers`. What about the default of `isgen`, which is present in the `Mela` class (https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/Mela.h#L67)?

MELA/interface/TVar.hh
```diff
@@ -26,7 +26,7 @@ typedef std::vector<SimpleParticle_t> SimpleParticleCollection_t;
 
 
 namespace TVar{
-  enum{
+  enum KAssociated{
```
usarica, line 29: These are internal flags, so I don't think they need an enum class structure. Naming should also not start with `K`.

MELA/python/mela_binding.cpp
```diff
@@ -1259,6 +1335,21 @@ PYBIND11_MODULE(Mela, m) {
...
+        .value("SMSyst_ResUp", TVar::SMSyst_ResUp)
+        .value("SMSyst_ResDown", TVar::SMSyst_ResDown);
+    
+    py::enum_<TVar::KAssociated>(m, "KAssociated")
+        .value("kNoAssociated", TVar::kNoAssociated)
+        .value("kUseAssociated_Leptons", TVar::kUseAssociated_Leptons)
+        .value("kUseAssociated_Photons", TVar::kUseAssociated_Photons)
+        .value("kUseAssociated_Jets", TVar::kUseAssociated_Jets)
+        .value("kUseAssociated_UnstableTops", TVar::kUseAssociated_UnstableTops)
+        .value("kUseAssociated_StableTops", TVar::kUseAssociated_StableTops);
```
usarica, line 1352: Same here. They should not be visible to the Python user.

usarica (changes requested):

MELA/python/mela_binding.cpp
MohitS704: Give me some time for this - in order to do this I have to fill out the implementation for `MELACandidate`, `MELAParticle`, and `MELAThreeBodyDecayCandidate.` I will make a commit after testing these implementations.

MELA/python/mela_binding.cpp
MohitS704: It does work for `None`! Good catch. The wrapper is now not necessary, one can pass None in. I can set the default parameters as well.

MELA/setup.sh
MohitS704: ZZGG for MCFM does not work otherwise in newer CMSSW made for el9. Is there a specific reason you are opposed to this? Are you concerned about something regarding the architectures?

MohitS704:
Okay so with this most recent push the following things were done

- The functions `const double& getMelaHiggsMass(int index=0)const` and `const double& getMelaHiggsWidth(int index=0)const;` have been added. I think they're necessary since `getPrimaryMass` and `getPrimaryWidth` don't necessarily correspond with the _actual_ Higgs mass or width in the matrix element. 
I think that they should maybe redirect to the mHiggs/wHiggs array if they're called, but that decision is also up to you @usarica 

- The default to `setInputEvent` was changed from 0 to nullptr, since it seems like that fits convention

- The debugging functions have been added as requested - they do not require an instantiation of a MELA object.
```cpp
    m.def("PrintCandidateSummary", [](TVar::simple_event_record curCand){
        TVar::simple_event_record* curRecord = &curCand;
        TUtil::PrintCandidateSummary(curRecord);
    });
    m.def("PrintCandidateSummary", [](MELACandidate curCand){
        MELACandidate* curRecord = &curCand;
        TUtil::PrintCandidateSummary(curRecord);
    });
```
- To implement the debugging functions, `MELACandidate`, `MELAParticle` and `MELAThreeBodyDecayCandidate` have been added to the Python binding (with proper inheritance from MELAParticle)

- I've given a few functions default arguments (i.e. `setInputEvent`

usarica:
@MohitS704 I think you need to keep commits separated, and perhaps even have different PRs to lump changes like these.

Referring to
https://github.com/JHUGen/JHUGenMELA/pull/51/files#diff-282f558c8b333c5116e51dc6206ea66acf38b63585ff8605c7c99ba73f96b65bR218-R224, you can already see that mass quantities are reset after the event is deleted, so there is not much utility to get the mass or width values for secondary resonances (keeping in mind that the user sets them to begin with, they should know what they are using). In practice, there is also extra redundancy wrt. the invariant mass of the MELACandidate either, so these functions do not bring much utility. They are also outside the scope of this PR as far as I am concerned.

usarica (changes requested):
Do not just revert [66ee630](https://github.com/JHUGen/JHUGenMELA/pull/51/commits/66ee6307bc4c8aded6f90e371fa615b29688b280) - remove it from the PR history, make a new commit for the rest of the changes, and make a new PR that contains the new functions you propose if you would still like to keep them. I will respond separately to the rest of the points.

MELA/setup.sh
usarica: Ok, I was just checking that we are not releasing new compiled libraries for no good reason.

MohitS704:
> Do not just revert [66ee630](https://github.com/JHUGen/JHUGenMELA/pull/51/commits/66ee6307bc4c8aded6f90e371fa615b29688b280) - remove it from the PR history, make a new commit for the rest of the changes, and make a new PR that contains the new functions you propose if you would still like to keep them. I will respond separately to the rest of the points.

This has been done @usarica

usarica (approved):

hroskes (approved):

PR #52, opened 2024-04-22
Madgraph Matrix Elements in MELA 

MohitS704:
With help from the rest of the JHU group (with a special mention to both Jeff and Michalis) I have interfaced with the madgraph matrix element reweighting tool for the onshell Z/Gamma* process. 

**My plan is to store the MADGRAPH subprocess areas within a repository inside this JHUGen organization.** For now, however, you can access the requisite materials [here](https://github.com/hexutils/madMELA_storage). There is one edit to the default FORTRAN code that I had to do - and that is evident [here](https://github.com/hexutils/madMELA_storage/blob/e0b58ed8bda33bebf58d522f2f4155e9179ca5aa/ggF-SMEFTsim-standalone_all/Source/MODEL/coupl.inc#L21) (I have changed the name `masses` to `mad_masses` so that the MCFM and MADGRAPH common blocks do not interfere).

Currently the only difference from the user-side is that you must change the matrix element to be `TVAR::MADGRAPH` and in `Mela::setInputEvent` you must set `madMela=true`. The couplings also work slightly differently, and are raw values (see the code [here](https://github.com/MohitS704/JHUGenMELA/blob/0b66d3f571c9aef1fea7e13bde23c97c6224c1b0/MELA/interface/Mela.h#L357).

You are able to edit Yukawa coupling masses with `resetYukawaMass` in `Mela.cc` [here](https://github.com/MohitS704/JHUGenMELA/blob/0b66d3f571c9aef1fea7e13bde23c97c6224c1b0/MELA/src/Mela.cc#L506).

I have opted to not make madMELA an object, as I think that would complicate a lot, seeing as how it would need to initialize many things that could not be shared between ZZMatrixElement and TUtils, etc. Instead, it is simply a namespace that can be accessed.

**This is obviously something that is in beta. I am sure that many of you will have feedback and opinions on the manner in which I implemented these changes, so please let me know what you think.**


MohitS704:
For full transparency, I am copying Andrei's email here to developers here:

> Dear colleagues, 
> 
> Mohit, with assistance from Jeff and initial contributions from Michalis, along with encouragement from several others, 
> has been working on the implementation of MadGraph matrix elements in MELA. The aim is to enhance the capability 
> of MELA to handle not only processes implemented in JHUGen and MCFM, or analytically, but potentially in other generators 
> as well. At present, this is merely a pilot project aimed at exploring the technical aspects, but progress appears promising. 
> Plots provided by Mohit demonstrate that both JHUGen and MadGraph yield essentially the same ratio of matrix elements 
> in the H->4l process, which can be utilized for either re-weighting or kinematic discriminants:
> 
> [https://msrivast.web.cern.ch/.totally_not_a_secret_folder/JHUGenMELA_testing/](https://nam02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmsrivast.web.cern.ch%2F.totally_not_a_secret_folder%2FJHUGenMELA_testing%2F&data=05%7C02%7Cmsrivas6%40jhu.edu%7C4d4a0bccf2514c77565b08dc62ddce39%7C9fa4f438b1e6473b803f86f8aedf0dec%7C0%7C0%7C638493951983444214%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C0%7C%7C%7C&sdata=wBjNDzVrr9v9gfohY9amOBjCsx%2BLetN5OC0q9PKQ2OU%3D&reserved=0)
> 
> Mohit has been invited to present the same MELA tutorial that he prepared for the LPC off-shell workshop at the LPC EFT 
> workshop in a few days. Consequently, he intends to introduce the proposal to incorporate additional matrix elements in MELA, 
> which could generate the interest of our EFT colleagues within CMS.
> 
> I expect that Mohit is ready to make a PR to show how the infrastructure works. The intent would be not to start using it officially, 
> but rather have a beta version to make sure that pieces are in the right places. 
> 
> Regards, 
> 
> Andrei

usarica (changes requested):
This will take some time to review the changes, so I am putting a preliminary review hold. I would not call these features as ready to be shown in a workshop, but I would certainly support the idea to access MG MEs using JHUGen.

gritsan:
We're not yet prepared to request people to use the code, but it's a good idea to showcase the concept and gather user interest. This is what the workshops are for.

MohitS704:
> This will take some time to review the changes, so I am putting a preliminary review hold. I would not call these features as ready to be shown in a workshop, but I would certainly support the idea to access MG MEs using JHUGen.

Hello @usarica, the MG MEs were simply teased at the EFT workshop. Mentioned it as being in development with a current beta for the Z/Gamma to 4l state to get some interest.

What is the status of your review hold?

usarica:
> > This will take some time to review the changes, so I am putting a preliminary review hold. I would not call these features as ready to be shown in a workshop, but I would certainly support the idea to access MG MEs using JHUGen.
> 
> Hello @usarica, the MG MEs were simply teased at the EFT workshop. Mentioned it as being in development with a current beta for the Z/Gamma to 4l state to get some interest.
> 
> What is the status of your review hold?

Good.

I will probably be able to take a shot at it over the weekend.

MohitS704:
Any thoughts @usarica?

usarica (changes requested):
A lot of cleanup is needed. It looks like there is also a lot of unnecessary duplication. As also mentioned, a global candidate is a bad idea for future parallelization purposes - in fact, for multiple MELACandidates stored in MELA, this would not be sensible.

Please make adjustments to the code as suggested, in a careful way, and ping the thread once you are done. I will then re-review the changes.

MELA/interface/Mela.h
```diff
@@ -347,6 +354,98 @@ class Mela{
   double selfDAZffcoupl[SIZE_AZff][2];
   // That is a lot of them!
 
+  //madMELA couplings
```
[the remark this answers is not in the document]
MohitS704: Are you sure this is the best way to do this? This would entail creating multiple different abstractions for a set of couplings that is not defined via a list and enumerations.

For example - looking at `MadMela.h` - one would have to create arrays for various warsaw operators (in some order), then create a structure to parse that array and link them to the original values in `params_r`, which, at the end of the day, is the same thing as a reference. Also, one would need the exact same number of enumerations as there are operators listed here in this block - rather than one space being cluttered - the user is forced to ping pong back and forth between enumerations, an array, and a common block.

I understand trying to keep things consistent, but I think the benefits of a reference outweigh the negatives - otherwise it adds many lines and complications for something that would effectively be done the same via a block of references.

MELA/makefile
```diff
@@ -43,7 +44,7 @@ ROOTINC = $(ROOFITINC)
 
 CC = g++
 CPPINC = -I$(MELADIR)/interface $(ROOTINC)
-CPPLOAD = -L$(MELALIBDIR) -lmcfm_709 -ljhugenmela -lcollier $(LIBS)
+CPPLOAD = -L$(MELALIBDIR) -lmcfm_709 -ljhugenmela -lcollier -L$(MADGRAPHLIBDIR) -lggF_SMEFTsim $(LIBS)
```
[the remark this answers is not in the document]
MohitS704: Currently only ggF is supported - that is the test case for the future - but I can always rename the library.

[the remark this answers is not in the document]
MohitS704: While it is certainly true that not everybody needs this functionality, the same could be said for MCFM. There are a lot of people who practically never use the MCFM matrix elements, yet are still required to compile MCFM. Given that we do not have a flag for MCFM, I don't see why we should have a flag for Madgraph.

MELA/python/mela_binding.cpp
```diff
@@ -218,6 +218,19 @@ SimpleParticleCollection_t collection_initializer(py::list listOfParticles){
...
+    }
+    if (associated->size() == 0){
+        associated = 0;
+    }
+    if (mothers->size() == 0){
+        mothers = 0;
+    }
+    mela.setInputEvent(daughters, associated, mothers, isgen, madMela);
+}
+
```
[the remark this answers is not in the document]
MohitS704: Neither am I lol - I think this is a branch merge holdover. My bad.

MELA/interface/Mela.h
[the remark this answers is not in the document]
usarica: > Are you sure this is the best way to do this? This would entail creating multiple different abstractions for a set of couplings that is not defined via a list and enumerations.
> 
> For example - looking at `MadMela.h` - one would have to create arrays for various warsaw operators (in some order), then create a structure to parse that array and link them to the original values in `params_r`, which, at the end of the day, is the same thing as a reference. Also, one would need the exact same number of enumerations as there are operators listed here in this block - rather than one space being cluttered - the user is forced to ping pong back and forth between enumerations, an array, and a common block.

Well, you can think of the self-defined spin zero couplings as an example. In those cases, we would be able to set `selfDHzzcoupl[resonance_index][coupling_enumerator][real_or_imaginary]=value'. What I am suggesting is a similar structure. You could communicate this structure for all couplings in MG.

> 
> I understand trying to keep things consistent, but I think the benefits of a reference outweigh the negatives - otherwise it adds many lines and complications for something that would effectively be done the same via a block of references.

MELA/makefile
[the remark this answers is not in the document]
usarica: Yes, if we are sure that this will be a single library, we should keep its name generic and leave it to the documentation to describe that only ggH couplings are supported.

MELA/python/mela_binding.cpp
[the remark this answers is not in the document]
usarica: Yes, but I think we were already considering to rewrite these lines and include checks for null pointers.

MELA/interface/MadMela.h
```diff
@@ -0,0 +1,70 @@
...
+        void smatrixhel_(int pdgs[], int& procid, const int& npdg, double* p, double& alphas, double& scale2, int& nhel, double& ans);
+        void update_all_coup_();
+    }
+    void setDefaultMadgraphValues();
+    void initialize_madMELA();
+
+    extern TVar::VerbosityLevel myVerbosity_;
+    const pair<const complex<double>&, const complex<double>&> getCKMMatEl(int i, int j);
+
+    extern madMelaInput* madMelaCandidate;
```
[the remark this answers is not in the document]
MohitS704: My understanding is that the biggest barrier to parallelization is the base FORTRAN - as the common blocks instantiate one space in memory that is always shared between subprocesses of the same compiled system. I actually have experimented with (and I think managed to get) some parallelization with MELA, but through a very complicated mess of compilation and a large headache.

Regardless, I see that there is the capability to have multiple MELACandidates - which is functionality I have never used personally. So yes, in that vein, I agree - I can try and interface with the MELACandidate system.

MELA/src/MadMela.cc
```diff
@@ -0,0 +1,565 @@
...
+#include "MELAStreamHelpers.hh"
+#include "TLorentzVector.h"
+#include "TUtil.hh"
+#include "TVar.hh"
+#include "TMCFM.hh"
+using MELAStreamHelpers::MELAout;
+using MELAStreamHelpers::MELAerr;
+
+madMela::madMelaInput* madMela::madMelaCandidate = nullptr;
+TVar::VerbosityLevel madMela::myVerbosity_ = TVar::SILENT;
```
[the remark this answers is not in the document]
MohitS704: What should I default to? Debug?

MELA/interface/MadMela.h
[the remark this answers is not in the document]
usarica: > My understanding is that the biggest barrier to parallelization is the base FORTRAN - as the common blocks instantiate one space in memory that is always shared between subprocesses of the same compiled system. I actually have experimented with (and I think managed to get) some parallelization with MELA, but through a very complicated mess of compilation and a large headache.

Yes, we will have to sort those issues, but that will need some restructuring of the lower-level code. It would still go more smoothly if the functionality exposed to the user is more lenient - otherwise, we would be discussing rewriting MELA altogether.

> 
> Regardless, I see that there is the capability to have multiple MELACandidates - which is functionality I have never used personally. So yes, in that vein, I agree - I can try and interface with the MELACandidate system.

MELA/src/MadMela.cc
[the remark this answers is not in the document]
usarica: TVar::ERROR (https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TVar.hh#L49).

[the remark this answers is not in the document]
usarica: Also the default of MELA: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/Mela.h#L42 (you might actually want to communicate the MELA flag directly).

MELA/interface/Mela.h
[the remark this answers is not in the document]
MohitS704: > Well, you can think of the self-defined spin zero couplings as an example. In those cases, we would be able to set `selfDHzzcoupl[resonance_index][coupling_enumerator][real_or_imaginary]=value'. What I am suggesting is a similar structure. You could communicate this structure for all couplings in MG.

This still feels like too much - especially given that all of these EFT terms are real valued. It feels much less natural for someone to say `Mela.mdl_ch = 1` than it does `Mela.MadgraphCouplings[gHiggs_ch] = 1`, because it's an unnecessary step in the process. 

Rather than the user assigning a value directly to the reference, we now have to:
1. create and instantiate an array to be passed with X coupling indices
2. create a map with X values between the enum and a reference to the common block to link the array and the MG operator values
3. create a function to assign each index to its reference

This seems very redundant for something that, in the end, will use the exact same reference procedure, just with more work and obfuscation.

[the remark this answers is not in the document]
usarica: > > Well, you can think of the self-defined spin zero couplings as an example. In those cases, we would be able to set `selfDHzzcoupl[resonance_index][coupling_enumerator][real_or_imaginary]=value'. What I am suggesting is a similar structure. You could communicate this structure for all couplings in MG.
> 
> This still feels like too much - especially given that all of these EFT terms are real valued. It feels much more natural for someone to say `Mela.mdl_ch = 1` than it does `Mela.MadgraphCouplings[gHiggs_ch] = 1`, because it's an unnecessary step in the process.

While I understand where you are coming from, i.e., the new Python implementation, the C++ interface is the primary one, and it needs to be well-maintainable. In that sense, adding coupling references to Mela is something to avoid. I should also note that the Python interface is not the only interface to Mela either, so how the different interfaces make it flexible to access the couplings from the original C++ implementation is a secondary matter of concern for the design of the C++ interface itself.

The fact that EFT terms are real-valued is also not a major concern - the struct could easily assume real-valued couplings (you just do not add the second index [2] in the array definition).

> 
> Rather than the user assigning a value directly to the reference, we now have to:
> 
> 1. create and instantiate an array to be passed with X coupling indices

That's just a matter of writing standard default and copy constructors for a coupling set, see for example https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TCouplings.cc#L12-L14.

> 2. create a map with X values between the enum and a reference to the common block to link the array and the MG operator values
> 3. create a function to assign each index to its reference

This is something you have to have done (in this case, in linking the references in Mela and those in the Fortran common block), so one does not need to write extra code for it any longer, aside from changing the location of the assignment in the code, and perhaps minor renaming as necessary. See for example https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L3360 for MCFM couplings.

References to the common block are not needed. As shown in the example, linking the Fortran library automatically makes the couplings available, in the format that is visible when executing `nm [library.so]`.

Please also note that if you have any booleans or plan to have them in the future, you need to be careful about how many bytes they end up consuming in the original Fortran implementation. For example, `noglue` is defined as a `bool` for MCFM: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TMCFM.hh#L269, but in this JHUGen function, https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TModJHUGenMELA.hh#L24, `usewwcoupl` is defined to have the type `int` even though the Fortran subroutine defines it as a `logical`, https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/fortran/mod_JHUGenMELA.F90#L130. Misusing the types could cause memory offsets that are not easy to track down (because they will look ok *most of the time*).

> 
> This seems very redundant for something that, in the end, will use the exact same reference procedure, just with more work and obfuscation.

MohitS704:
@usarica Is there a function that generates mother gluons for you if they are not provided? The Madgraph MEs require gluons, and I was wondering if such functionality already existed.

MELA/interface/Mela.h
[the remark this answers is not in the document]
MohitS704: > While I understand where you are coming from, i.e., the new Python implementation, the C++ interface is the primary one, and it needs to be well-maintainable. In that sense, adding coupling references to Mela is something to avoid. I should also note that the Python interface is not the only interface to Mela either, so how the different interfaces make it flexible to access the couplings from the original C++ implementation is a secondary matter of concern for the design of the C++ interface itself.

The implementation of the Python bindings are agnostic to the way that the C++ couplings are done - they would be there regardless of how the C++ is implemented. In fact, it would probably be easier for the Python bindings to operate via the arrays since those are tried and tested. My issues lie with a perceived incongruity between the stated design philosophy and the adherence to previous structure when (I believe it to be) unnecessary.

> The fact that EFT terms are real-valued is also not a major concern - the struct could easily assume real-valued couplings (you just do not add the second index [2] in the array definition).

Yes fair enough - but my point here was more to state that a new structure is unnecessary, imo.
 
> That's just a matter of writing standard default and copy constructors for a coupling set, see for example https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TCouplings.cc#L12-L14.

I think we're getting closer to something here. My issue does not lie with any interface, but rather that I believe the nature of the EFT terms do not lend themselves to an array - and I think that making a new object solely for the purposes of assigning terms is unnecessary. In vein with the design philosophies you were saying earlier, I think it makes for very redundant code!

> > 2. create a map with X values between the enum and a reference to the common block to link the array and the MG operator values
> > 3. create a function to assign each index to its reference
> 
> This is something you have to have done (in this case, in linking the references in Mela and those in the Fortran common block), so one does not need to write extra code for it any longer, aside from changing the location of the assignment in the code, and perhaps minor renaming as necessary. See for example https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L3360 for MCFM couplings.

Except in this case, I would have to here, right? I would have to generate a new array in the SpinZeroCouplings class, then write a SpinZeroCouplings method to assign each array element (let's call it `SetEFTCouplings`), and write a bunch of enums to do so. It's a bunch of ping ponging around that involves writing much more code that is not self-explanatory (i.e. passing an array of enumerated size to a function in a class that then takes the enumerations and assigns terms to global variables), rather than simply making a block of reference assignments (that I am more than happy to move to the source file).

> References to the common block are not needed. As shown in the example, linking the Fortran library automatically makes the couplings available, in the format that is visible when executing `nm [library.so]`.

When I say "reference the common block" I mean reference the variables in the extern "C" block within the header file. I understand that they are available upon compilation from the original .so file. What I meant here was that every index in any array would need to be manually linked to the global variables in the extern block anyways - so why obfuscate where it's done?

> Please also note that if you have any booleans or plan to have them in the future, you need to be careful about how many bytes they end up consuming in the original Fortran implementation. For example, `noglue` is defined as a `bool` for MCFM: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TMCFM.hh#L269, but in this JHUGen function, https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/interface/TModJHUGenMELA.hh#L24, `usewwcoupl` is defined to have the type `int` even though the Fortran subroutine defines it as a `logical`, https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/fortran/mod_JHUGenMELA.F90#L130. Misusing the types could cause memory offsets that are not easy to track down (because they will look ok _most of the time_).

Yes - As I understand from [here](http://www.yolinux.com/TUTORIALS/LinuxTutorialMixingFortranAndC.html), this is true - and good advice!

MELA/makefile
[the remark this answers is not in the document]
MohitS704: > Yes, if we are sure that this will be a single library, we should keep its name generic and leave it to the documentation to describe that only ggH couplings are supported.

Actually I am unsure if this will be a single library. If we add a different process, say VBS for example, I'm not sure if that would be encompassed in the ggF library or a different one. I guess for now we can assume all processes can be subsumed into one with the renaming.

MELA/python/mela_binding.cpp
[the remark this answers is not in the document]
MohitS704: > Yes, but I think we were already considering to rewrite these lines and include checks for null pointers.

Yeah, I did do that - this is just some stuff where some changes weren't properly propagated to the new branch.

usarica:
> @usarica Is there a function that generates mother gluons for you if they are not provided? The Madgraph MEs require gluons, and I was wondering if such functionality already existed.

Indeed, there is a function `TUtil::GetBoostedParticleVectors` that extracts the plain particle ids and momenta from a given `MELACandidate` based on the specified requirements of the ME calculation: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L7947. There is no function that just extracts momenta for mothers as the extraction in the general setting depends on the number of requested outgoing particles associated with Higgs production, so we perform the interpretation of the candidate object in one go.

The note on how mothers are treated are in these comment blocks: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L8343-L8347.

Here is an example use case that you might find helpful for a LO ggH calculation: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L4808-L4815.

Here is also an analogous example from MCFM calls: https://github.com/JHUGen/JHUGenMELA/blob/master/MELA/src/TUtil.cc#L4112-L4162

MELA/interface/Mela.h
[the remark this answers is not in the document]
usarica: Hmm, let me also approach from a different direction then: Is there a non-degeneracy between the couplings already in JHUGen and those that are introduced for MG? Given that you mention these are spin-0 couplings, can't we simply translate JHUGen/MCFM couplings, like how it is done in the Lexicon? In fact, it could be easier to do here than in Lexicon, given that we already have access to all of the couplings/masses from extern bindings.

> It's a bunch of ping ponging around that involves writing much more code that is not self-explanatory (i.e. passing an array of enumerated size to a function in a class that then takes the enumerations and assigns terms to global variables), rather than simply making a block of reference assignments (that I am more than happy to move to the source file).

The problem with this approach would be later in implementing parallelization. We will for sure have to figure out how to handle the Fortran component either, since we would not want a serialization bottleneck right in the middle of the ME calculation workflow, but another serialization bottleneck would be introduced if we do not compartmentalize users' specification of couplings and the assignment of the values of extern-referenced couplings in the ME libraries. The current implementation was designed to be able to prevent this secondary bottleneck in the future.

Let me give you also the model that was ultimately in mind in the design: One could think of a parallelization model where we have a collection of `C(Nevents, NMEs, Ncouplings)` couplings (in ordinary event processing, `C(:, :, k)` have the same values for a given `k`, but different use cases do not have to keep the same couplings), and `I(Nevents, NMEs, Nreco)` input candidates. Then, one could write a workflow that computes the collection of MEs `F(Nevents, NMEs, Nreco)[C, I]`. Currently, the coupling structures allow us to collect `C(Nevents, NMEs, Ncouplings)` in a structured way as `S(Nevents, NMEs)[C]`, allowing us to write `F(Nevents, NMEs, Nreco)[S[C], I]`. Computations over the rest of the indices happen serially as the user executes the ME calls at the moment, but given that we have the structure to collect many input candidates, we could add the feature for batch computation, either over the flattened vector of inputs `I` as it is currently implemented (as a 1D `vector`), or still in series for `(Nevents, NMEs)`, as the user would like.

What you propose breaks the structuring `S[C]` explicitly, introducing a further complication down the road, and I would prefer to avoid that. It is not just a consistency matter with the implementation of the rest of couplings; this is the reason why we preserved coupling structures even though for serial processing, it might seem like introducing an unnecessary layer of complication.

Does what I propose require implementation of batch processing capabilities? Yes, but the current implementation reduces that to a matter of passing batches of couplings and ME specifications, and returning a batch of numbers at the end of the call. Internally, we would have to figure out how to parallelize the Fortran code, which is where I would suggest to really focus our energy.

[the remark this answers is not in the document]
MohitS704: > Hmm, let me also approach from a different direction then: Is there a non-degeneracy between the couplings already in JHUGen and those that are introduced for MG? Given that you mention these are spin-0 couplings, can't we simply translate JHUGen/MCFM couplings, like how it is done in the Lexicon? In fact, it could be easier to do here than in Lexicon, given that we already have access to all of the couplings/masses from extern bindings.

Currently I would hold off on the Lexicon translations - Lexicon doesn't cover every operator here - and I don't really expect it to anytime soon.

> > It's a bunch of ping ponging around that involves writing much more code that is not self-explanatory (i.e. passing an array of enumerated size to a function in a class that then takes the enumerations and assigns terms to global variables), rather than simply making a block of reference assignments (that I am more than happy to move to the source file).
> 
> The problem with this approach would be later in implementing parallelization. We will for sure have to figure out how to handle the Fortran component either, since we would not want a serialization bottleneck right in the middle of the ME calculation workflow, but another serialization bottleneck would be introduced if we do not compartmentalize users' specification of couplings and the assignment of the values of extern-referenced couplings in the ME libraries. The current implementation was designed to be able to prevent this secondary bottleneck in the future.
> 
> Let me give you also the model that was ultimately in mind in the design: One could think of a parallelization model where we have a collection of `C(Nevents, NMEs, Ncouplings)` couplings (in ordinary event processing, `C(:, :, k)` have the same values for a given `k`, but different use cases do not have to keep the same couplings), and `I(Nevents, NMEs, Nreco)` input candidates. Then, one could write a workflow that computes the collection of MEs `F(Nevents, NMEs, Nreco)[C, I]`. Currently, the coupling structures allow us to collect `C(Nevents, NMEs, Ncouplings)` in a structured way as `S(Nevents, NMEs)[C]`, allowing us to write `F(Nevents, NMEs, Nreco)[S[C], I]`. Computations over the rest of the indices happen serially as the user executes the ME calls at the moment, but given that we have the structure to collect many input candidates, we could add the feature for batch computation, either over the flattened vector of inputs `I` as it is currently implemented (as a 1D `vector`), or still in series for `(Nevents, NMEs)`, as the user would like.
> 
> What you propose breaks the structuring `S[C]` explicitly, introducing a further complication down the road, and I would prefer to avoid that. It is not just a consistency matter with the implementation of the rest of couplings; this is the reason why we preserved coupling structures even though for serial processing, it might seem like introducing an unnecessary layer of complication.
> 
> Does what I propose require implementation of batch processing capabilities? Yes, but the current implementation reduces that to a matter of passing batches of couplings and ME specifications, and returning a batch of numbers at the end of the call. Internally, we would have to figure out how to parallelize the Fortran code, which is where I would suggest to really focus our energy.

Ah okay - now it makes much more sense why we do this. I think it would be best if I went down the SpinZeroCouplings route for now. In the future (AKA once we have the madMELA fleshed out for more processes + maybe offshell, etc.) we can start working to implement Lexicon's language into MELA itself - that seems like a good task.

I'm glad we worked this through!

MELA/interface/MadMela.h
```diff
@@ -0,0 +1,70 @@
+#ifndef madMELA
+#define madMELA
+
+#include <complex>
+#include <map>
+#include <vector>
+#include "TVar.hh"
+using namespace std;
```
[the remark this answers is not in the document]
MohitS704: the complex is needed for the couplings as a datatype - the rest have been removed.

MELA/src/MadMela.cc
```diff
@@ -0,0 +1,565 @@
...
+void madMela::set_wHiggs(double myHiggsWidth, int index){
+    if((index < 0) || (index > 1)){
+        throw out_of_range("Higgs width index can only be 1 or 0!");
+    }
+    else if(index == 0){
+        madMela::widths_.mdl_wh = myHiggsWidth;
+    } else{
+        madMela::widths_.mdl_wh1 = myHiggsWidth;
+    }
+}
```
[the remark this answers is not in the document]
MohitS704: Functionality moved to the TUtil functions for mass and width

MELA/src/MadMela.cc
```diff
@@ -0,0 +1,565 @@
...
+//     *matElAndConjugate.second = conj(matEl);
+// }
+
+// const pair<const complex<double>&, const complex<double>&> madMela::getCKMMatEl(int i, int j) const{
+//     pair<complex<double>*, complex<double>*> matElAndConjugate= madMela::CKM_map(<int, int>(i, j));
+
+//     const complex<double>& matEl = as_const(*matElAndConjugate.first);
+//     const complex<double>& Conjugate = as_const(*matElAndConjugate.second);
+//     return const pair<const complex<double>&, const complex<double>&>(matEl, Conjugate);
+// }
```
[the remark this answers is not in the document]
MohitS704: done.

MELA/src/MadMela.cc
```diff
@@ -0,0 +1,565 @@
...
+        i++;
+    }
+
+    madMela::madMelaCandidate = new madMela::madMelaInput(pdgs, p);
+    if(delete_associated){
+        delete pAssociated;
+    }
+}
+
+void madMela::computeP(double& prob, int nhel){
```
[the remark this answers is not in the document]
MohitS704: Added function `TUtil::MadgraphMatEl`

MELA/src/MadMela.cc
```diff
@@ -0,0 +1,565 @@
...
+    madMela::params_r_.mdl_ymtau = TUtil::GetMass(15);
+
+    madMela::params_r_.mdl_aew = 7.815553e-03;
+    madMela::params_r_.mdl_gf = ewcouple_.Gf; //1.166379e-05;
+    madMela::params_r_.as = qcdcouple_.as; //1.179000e-01;
+
+    madMela::setDefaultMadgraphValues();
+}
+
+void madMela::setInputEvent(SimpleParticleCollection_t* pDaughters, SimpleParticleCollection_t* pAssociated, SimpleParticleCollection_t* pMothers, bool isGen){
```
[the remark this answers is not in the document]
MohitS704: Done.

MELA/interface/MadMela.h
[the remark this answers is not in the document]
MohitS704: Done.

MELA/src/Mela.cc
```diff
@@ -301,16 +328,28 @@ void Mela::setInputEvent(
...
-    pDaughters,
-    pAssociated,
-    pMothers,
-    isGen
-    );
+  pDaughters,
+  pAssociated,
+  pMothers,
+  isGen
+  );
```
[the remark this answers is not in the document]
MohitS704: Done.

MELA/src/MadMela.cc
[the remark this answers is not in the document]
MohitS704: MadMELA doesn't have a verbosity anymore - not necessary as all other functions handle things that already have verbosity values.

MELA/makefile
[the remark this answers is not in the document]
MohitS704: renamed the library to `libSMEFTsim.so`. Any future libraries should attempt to be merged into this one through diagram filters or utilization of .o files.

MELA/interface/Mela.h
```diff
@@ -199,7 +205,8 @@ class Mela{
     );
   void computeP(
     float& prob,
-    bool useConstant=true
+    bool useConstant=true,
+    int nhel=-1
```
MohitS704, line 209: nhel is added as a default parameter (-1 implies a sum over all helicities a la JHUGenMELA) in order to allow for someone to choose a helicity value (as you are able to do so in Madgraph reweighting). If madgraph reweighting is not used, nhel will never be accessed.

MELA/interface/Mela.h
```diff
@@ -345,7 +352,10 @@ class Mela{
   double selfDaTQGCcoupl[SIZE_ATQGC][2];
   //****Anomnalous Zff**//
   double selfDAZffcoupl[SIZE_AZff][2];
-  // That is a lot of them!
+  //****SMEFTSIM terms**//
+  double selfDSmeftSimcoupl[SIZE_SMEFT]; //madMela
```
MohitS704, line 356: selfDSmeftSimcoupl stores the SMEFTsim couplings through `TUtil::SetMadgraphSpinZeroCouplings` and the `CouplingIndex_Smeft` enum in `TCouplingsBase.hh`

MELA/interface/Mela.h
usarica: I would prefer to have that as a property of the particle, i.e. an additional property in MELAParticle, which would get inherited by all daughter classes. Moreover, helicities are particle properties, not a general total matrix element property. To also add, -1 could be confusing as 'down' - it might be better to assign -99 as a default. To avoid consuming unnecessary memory, it is best to keep the type minimal ([signed] char).

MELA/interface/MadMela.h
```diff
@@ -0,0 +1,54 @@
+#ifndef madMELA
+#define madMELA
+
+#include <complex>
+using namespace std;
```
usarica, line 5: Remove std.

MELA/interface/MadMela.h
```diff
@@ -0,0 +1,54 @@
...
+            mdl_chq3,mdl_chu,mdl_chd,mdl_chudre,mdl_cll,mdl_cll1,mdl_cqq1,mdl_cqq11,mdl_cqq3,mdl_cqq31,mdl_clq1,mdl_clq3,mdl_cee,mdl_cuu,mdl_cuu1,mdl_cdd,mdl_cdd1,mdl_ceu,mdl_ced,mdl_cud1,mdl_cud8,mdl_cle,mdl_clu,mdl_cld,mdl_cqe,
+            mdl_cqu1,mdl_cqu8,mdl_cqd1,mdl_cqd8,mdl_cledqre,mdl_cquqd1re,mdl_cquqd11re,mdl_cquqd8re,mdl_cquqd81re,mdl_clequ1re,mdl_clequ3re,mdl_cgtil,mdl_cwtil,mdl_chgtil,mdl_chwtil,mdl_chbtil,mdl_chwbtil,mdl_cewim,mdl_cebim,
+            mdl_cugim,mdl_cuwim,mdl_cubim,mdl_cdgim,mdl_cdwim,mdl_cdbim,mdl_chudim,mdl_cehim,mdl_cuhim,mdl_cdhim,mdl_cledqim,mdl_cquqd1im,mdl_cquqd8im,mdl_cquqd11im,mdl_cquqd81im,mdl_clequ1im,mdl_clequ3im,mdl_lambdasmeft,mdl_aew,
+            mdl_gf,as,mdl_linearpropcorrections,mdl_ymdo,mdl_ymup,mdl_yms,mdl_ymc,mdl_ymb,mdl_ymt,mdl_yme,mdl_ymm,mdl_ymtau;
+        }params_r_;
+
+        extern struct{
+            complex<double> mdl_g__exp__3,mdl_complexi,mdl_ceh,mdl_cuh,mdl_cdh,mdl_cew,mdl_ceb,mdl_cug,mdl_cuw,mdl_cub,mdl_cdg,mdl_cdw,mdl_cdb,mdl_chud,mdl_cledq,mdl_cquqd1,mdl_cquqd11,mdl_cquqd8,mdl_cquqd81,mdl_clequ1,mdl_clequ3,
```
usarica, line 31: Are you sure the data structure is not representable with a simple array of two elements? While the underlying memory outline might appear the same, there is no guarantee that specific implementations should respect that, is there?

MELA/interface/TCouplingsBase.hh
```diff
@@ -197,6 +197,93 @@ namespace{
 
     SIZE_AZff
   };
+  enum CouplingIndex_Smeft{
```
usarica, line 200: 'SMEFT' everywhere - let's avoid making the user think about capital vs small case letter differences.

MELA/src/Mela.cc
```diff
@@ -130,7 +137,7 @@ void Mela::cleanLinkedFiles() {
 
 
 void Mela::build(double mh_){
-  if (myVerbosity_>=TVar::DEBUG) MELAout << "Start Mela::build" << endl;
```
usarica, line 133: Line should remain here.

MELA/src/TEvtProb.cc
```diff
@@ -407,12 +411,11 @@ bool TEvtProb::CheckInputPresent(){
 
 // ME functions
 // Cross-section calculations for H + 0 jet
-double TEvtProb::XsecCalc_XVV(){
+double TEvtProb::XsecCalc_XVV(int nhel){
```
usarica, line 414: Same nhel commentary as in the other case.

MELA/src/TUtil.cc
```diff
@@ -5,7 +5,9 @@
 #include <utility>
 #include <algorithm>
 #include <cassert>
+#include <complex>
```
usarica, line 8: Avoid using C++ STL for C-style Fortran interfacing.

MELA/src/TUtil.cc
```diff
@@ -8428,22 +8702,28 @@ MELACandidate* TUtil::ConvertVectorFormat(
   std::vector<MELAParticle*> daughters;
   std::vector<MELAParticle*> aparticles;
   std::vector<MELAParticle*> mothers;
+  // MELAout << "daughters" << endl;
```
usarica, line 8705: Remove added comment lines.

usarica (changes requested):
Changes needed have been indicted in individual comments.

MELA/interface/Mela.h
MohitS704: The helicity isn't actually a property of the particle in this case - it's an input to the reweighting function for Madgraph. The `smatrixhel` function takes in a list of pdgs, process IDs, pdg numbers, 4 vectors, an alphaS value, a scale2 value (irrelevant here), the helicity number (-1 if summing over all helicities), and the answer.

MELA/src/TUtil.cc
MohitS704: The complex here is for the `getCKM` function, not for Fortran interfacing.

MELA/interface/MadMela.h
MohitS704: To be honest I have not tried representing it this way - however it does work well since it's interfacing with C++ directly in functions like `getMadgraphCKMElement'.

MohitS704: I've seen nothing explicitly warning against this in practice, and I've tested it to work, which is why I've kept it in.

MohitS704:
I've either made the changes or explained my reasoning for why I did what I did above.

MELA/interface/MadMela.h
usarica: There wouldn't be a warning actually - it would just read other bytes in memory. What you do is fine for the GCC compiler, but it is not standardized and is therefore not portable (just adding a single boolean as a data member to complex is enough to cause an offset).

MELA/src/TUtil.cc
usarica: Yes, but the underlying returned objects are still from complex-type couplings.

usarica (changes requested):
Apart from the raised portability issue, the PR is ok.

MELA/src/TUtil.cc
MohitS704: I've left this complex include here, as the function now converts the array of size 2 into a complex number. 

I believe this allows for a more seamless integration into C++ as a getter, where one can simple get a value with the function then do any complex-valued operations without any more work.

MohitS704:
Okay, I think everything should be done here.

MELA/interface/MadMela.h
```diff
@@ -0,0 +1,53 @@
...
+            mdl_chq3,mdl_chu,mdl_chd,mdl_chudre,mdl_cll,mdl_cll1,mdl_cqq1,mdl_cqq11,mdl_cqq3,mdl_cqq31,mdl_clq1,mdl_clq3,mdl_cee,mdl_cuu,mdl_cuu1,mdl_cdd,mdl_cdd1,mdl_ceu,mdl_ced,mdl_cud1,mdl_cud8,mdl_cle,mdl_clu,mdl_cld,mdl_cqe,
+            mdl_cqu1,mdl_cqu8,mdl_cqd1,mdl_cqd8,mdl_cledqre,mdl_cquqd1re,mdl_cquqd11re,mdl_cquqd8re,mdl_cquqd81re,mdl_clequ1re,mdl_clequ3re,mdl_cgtil,mdl_cwtil,mdl_chgtil,mdl_chwtil,mdl_chbtil,mdl_chwbtil,mdl_cewim,mdl_cebim,
+            mdl_cugim,mdl_cuwim,mdl_cubim,mdl_cdgim,mdl_cdwim,mdl_cdbim,mdl_chudim,mdl_cehim,mdl_cuhim,mdl_cdhim,mdl_cledqim,mdl_cquqd1im,mdl_cquqd8im,mdl_cquqd11im,mdl_cquqd81im,mdl_clequ1im,mdl_clequ3im,mdl_lambdasmeft,mdl_aew,
+            mdl_gf,as,mdl_linearpropcorrections,mdl_ymdo,mdl_ymup,mdl_yms,mdl_ymc,mdl_ymb,mdl_ymt,mdl_yme,mdl_ymm,mdl_ymtau;
+        }params_r_;
+
+        extern struct{
+            std::array<double,2> mdl_g__exp__3,mdl_complexi,mdl_ceh,mdl_cuh,mdl_cdh,mdl_cew,mdl_ceb,mdl_cug,mdl_cuw,mdl_cub,mdl_cdg,mdl_cdw,mdl_cdb,mdl_chud,mdl_cledq,mdl_cquqd1,mdl_cquqd11,mdl_cquqd8,mdl_cquqd81,mdl_clequ1,mdl_clequ3,
```
usarica, line 30: Please add a static_assert to ensure memory layout is the same, or remove C++ containers in linking to Fortran code!

usarica: Same in other places.

MELA/interface/TCouplings.hh
```diff
@@ -23,6 +23,7 @@ public:
 
   void SetHVVpCouplings(unsigned int index, double c_real, double c_imag, bool setWWp = false, int whichResonance=1);
   void SetHVpVpCouplings(unsigned int index, double c_real, double c_imag, bool setWpWp = false, int whichResonance=1);
+  void SetSmeftSimCouplings(unsigned int index, double value);
```
usarica, line 26: Naming convention, should use 'SMEFT' consistently.

MELA/interface/TCouplings.hh
```diff
@@ -59,6 +60,8 @@ public:
   double Hwwpcoupl[SIZE_HVV][2];
   double Hwpwpcoupl[SIZE_HVV][2];
 
+  double SmeftSimcoupl[SIZE_SMEFT];
```
usarica, line 63: Here, too.

MELA/interface/TEvtProb.hh
```diff
@@ -97,7 +97,9 @@ public:
   // Reset the IO record, called at te beginning of each comoputation
   void ResetIORecord();
 
-  double XsecCalc_XVV();
+  double XsecCalc_XVV(
+    int nhel=-1
```
usarica, line 101: I think the question but forgot whether a good answer was provided. Wouldn't a default of -1 be confusing? What are other allowed values?

MELA/interface/TUtil.hh
```diff
@@ -207,7 +208,9 @@ namespace TUtil{
   void GetMassWidth(int ipart, double& m, double& ga);
   void GetMassWidth(const MELAParticle* part, double& m, double& ga);
   void SetCKMElements(double* invckm_ud, double* invckm_us, double* invckm_cd, double* invckm_cs, double* invckm_ts, double* invckm_tb, double* invckm_ub=0, double* invckm_cb=0, double* invckm_td=0);
+  void SetMadgraphCKMElements(double ckmlambda=-1, double ckma=-1, double ckmrho=-1, double ckmeta=-1);
```
usarica, line 211: 1) What is the reason to prefer the Wolfenstein parametrization instead of the standard?
2) What is the utility to pick a nonzero eta?
3) Why are the defaults all at -1 and not 0?

usarica: Regarding (2), what I am really asking is whether complex CKM element are needed at all.

MELA/interface/ZZMatrixElement.h
```diff
@@ -20,7 +20,8 @@ class ZZMatrixElement{
   ~ZZMatrixElement();
 
   void computeXS(
-    float &mevalue
+    float &mevalue,
+    int nhel=-1
```
usarica, line 24: Again, why this default?

MELA/interface/ZZMatrixElement.h
```diff
@@ -97,6 +98,7 @@ class ZZMatrixElement{
     double selfDHwwLambda_qsq[nSupportedHiggses][SIZE_HVV_LAMBDAQSQ][SIZE_HVV_CQSQ],
     int selfDHzzCLambda_qsq[nSupportedHiggses][SIZE_HVV_CQSQ],
     int selfDHwwCLambda_qsq[nSupportedHiggses][SIZE_HVV_CQSQ],
+    double selfDSmeftSimcoupl[SIZE_SMEFT],
```
usarica, line 101: Name...

MELA/src/Mela.cc
```diff
@@ -220,15 +228,16 @@ void Mela::build(double mh_){
   super->SetPathToCards(cardfile.substr(0, cardfile.length()-14).c_str());
   super->SetVerbosity((myVerbosity_>=TVar::DEBUG));
   super->init();
-
-
```
usarica, line 224: Keep at least one space for readability.

MELA/src/Mela.cc
```diff
@@ -220,15 +228,16 @@ void Mela::build(double mh_){
...
   /***** SuperDijetMela *****/
   float superDijetSqrts = LHCsqrts;
   if (superDijetSqrts<13.) superDijetSqrts=13.; // Dijet resolution does not exist for 7 or 8 TeV
   superDijet = new SuperDijetMela(superDijetSqrts, myVerbosity_);
 
   // Initialize the couplings to 0 and end Mela constructor
   reset_SelfDCouplings();
+  if (myVerbosity_>=TVar::DEBUG) MELAout << "Start Mela::build" << endl;
+  if(myVerbosity_>=TVar::DEBUG) MELAout << "Running madMela::initialize_madMELA" << endl;
```
usarica, line 239: Space.

MELA/src/Mela.cc
```diff
@@ -277,10 +289,14 @@ void Mela::setProcess(TVar::Process myModel, TVar::MatrixElement myME, TVar::Pro
     else if (myProduction_==TVar::JJQCD_S) myProduction_=TVar::JJQCD;
   }
   myModel_ = myModel;
+  if(myME_==TVar::MADGRAPH && myProduction_ != TVar::ZZGG && myProduction_ != TVar::ZZINDEPENDENT){
```
usarica, line 292: Space

MELA/src/Mela.cc
```diff
@@ -290,9 +306,15 @@ void Mela::setVerbosity(TVar::VerbosityLevel verbosity_){
...
-void Mela::setMelaHiggsMassWidth(double myHiggsMass, double myHiggsWidth, int index){ ZZME->set_mHiggs_wHiggs(myHiggsMass, myHiggsWidth, index); }
+void Mela::setMelaHiggsMass(double myHiggsMass, int index){
+  ZZME->set_mHiggs(myHiggsMass, index); 
+}
+void Mela::setMelaHiggsWidth(double myHiggsWidth, int index){
+  ZZME->set_wHiggs(myHiggsWidth, index); 
+}
+void Mela::setMelaHiggsMassWidth(double myHiggsMass, double myHiggsWidth, int index){
+  ZZME->set_mHiggs_wHiggs(myHiggsMass, myHiggsWidth, index); 
+}
```
usarica, line 317: Revert this change that does nothing.

MELA/src/Mela.cc
```diff
@@ -308,9 +330,12 @@ void Mela::setInputEvent(
...
     pMothers,
     isGen
-    );
+  );
 }
-void Mela::resetInputEvent(){ ZZME->reset_InputEvent(); }
+void Mela::resetInputEvent(){
+  madMela::setDefaultMadgraphValues();
+  ZZME->reset_InputEvent();
+  }
```
usarica, line 338: Space

MELA/src/Mela.cc
```diff
@@ -473,6 +518,13 @@ double Mela::getPrimaryMass(int ipart){ return ZZME->get_PrimaryMass(ipart); }
 double Mela::getPrimaryWidth(int ipart){ return ZZME->get_PrimaryWidth(ipart); }
 double Mela::getHiggsWidthAtPoleMass(double mass){ return ZZME->get_HiggsWidthAtPoleMass(mass); }
 
+void Mela::SetMadgraphCKMElements(double ckmlambda, double ckma, double ckmrho, double ckmeta){
+  TUtil::SetMadgraphCKMElements(ckmlambda, ckma, ckmrho, ckmeta);
+}
+complex<double> Mela::GetMadgraphCKMElement(int iquark, int jquark){
```
usarica, line 524: std::complex (should match style in header)

MELA/src/TUtil.cc
```diff
@@ -1522,9 +1528,65 @@ void TUtil::SetCKMElements(double* invckm_ud, double* invckm_us, double* invckm_
...
+  if(ckma >= 0){
+    madMela::params_r_.mdl_ckma = ckma;
+  }
+  if(ckmrho >= 0){
+    madMela::params_r_.mdl_ckmrho = ckmrho;
+  }
+  if(ckmeta >= 0){
+    madMela::params_r_.mdl_ckmeta = ckmeta;
+  }
+  madMela::update_all_coup_();
```
usarica, line 1544: Does this call really update all couplings? If so, isn't it better to move to a later stage when the ME is actually called (like calls to chooser in MCFM)?

MELA/src/TUtil.cc
```diff
@@ -1522,9 +1528,65 @@ void TUtil::SetCKMElements(double* invckm_ud, double* invckm_us, double* invckm_
...
+        return std::complex<double>(madMela::params_c_.mdl_ckm3x3[0], madMela::params_c_.mdl_ckm3x3[1]);
+      } else{
+        MELAerr << "TUtil::GetMadgraphCKMElement: Invalid second index!" << endl;
+      }
+      break;
+    default:
+      MELAerr << "TUtil::GetMadgraphCKMElement: Invalid first index!" << endl;
+      break;
+  }
+  return complex<double>(0);
```
usarica, line 1588: std::

MELA/src/TUtil.cc
```diff
@@ -1522,9 +1528,65 @@ void TUtil::SetCKMElements(double* invckm_ud, double* invckm_us, double* invckm_
...
+  }
+  madMela::update_all_coup_();
+}
 double TUtil::GetCKMElement(int iquark, int jquark){
   return __modparameters_MOD_ckmbare(&iquark, &jquark);
 }
+complex<double> TUtil::GetMadgraphCKMElement(int iquark, int jquark){
+  switch (iquark){
+    case 1:
+      if(jquark == 1){
```
usarica, line 1552: Code spacing style should be consistent.

MELA/src/TUtil.cc
```diff
@@ -4857,6 +5005,7 @@ double TUtil::JHUGenMatEl(
       p4[arrindex+2][3] = momTmp->Z()*GeV;
       MomStore[arrindex+2] = *momTmp;
       if (verbosity >= TVar::DEBUG) MELAout << "MYIDUP_tmp[" << arrindex << "(" << ipar << ")" << "]=" << MYIDUP_tmp[arrindex] << endl;
+      MELAout << "Mother" << ipar << " = " << p4[ipar][1] << " " << p4[ipar][2] << " " << p4[ipar][3] << " " << p4[ipar][4] << " " << endl;
```
usarica, line 5008: Please remove.

MELA/src/TUtil.cc
usarica, line 8705: Please remove.

usarica: Also the rest of these debugging lines.

usarica (changes requested):

MELA/interface/MadMela.h
MohitS704: I mean sure - but for an std::array it seems a little bit redundant, no? An std::array is quite literally just a wrapper around the normal C-style array.

usarica: Yes, the most logical implementation is a simple struct that wraps around a C-style array, but the Fortran interface needs to ensure that that logical implementation is indeed the case. Note that a certain style of implementation is not guaranteed by the C++ standard; this is the case for complex, and it is also the case for array, vector etc. I can come up with ways to implement the array object that has CUDA portability, for example, and it would not conform to the expectation aforementioned while still keeping contiguous memory storage.

This is why I asked you to avoid using C++ classes for Fortran interface, which would be better in C style to ensure the memory layout match.

MELA/interface/TEvtProb.hh
MohitS704: A value of -1 for the helicity indicates that you would like your matrix element to sum over all helicities, which is why it is the default. Any other value indicates a desired helicity value.

MohitS704: This is the setup for the input to the Madgraph calculation function, `smatrixhel`.

MELA/interface/TUtil.hh
MohitS704: 1. This is using a function that Madgraph uses to change its ckm matrix values. Direct editing is not possible, as the values would be overridden anyways through the `update_all_params` function.
2. Is there any reason to limit the user? If they never choose to do it they won't. 
   * as an aside, these are the default SM values listed in the parameter card. I could make those the default.
   ```Block ckmblock 
    2 2.265000e-01 # CKMlambda 
    3 7.900000e-01 # CKMA 
    4 1.410000e-01 # CKMrho 
    5 3.570000e-01 # CKMeta 
   ```
3. I can make the defaults the normal SM value from the datacard.

MELA/interface/ZZMatrixElement.h
MohitS704: See previous answer about inputs to the `smatrixhel` function.

MELA/src/TUtil.cc
MohitS704: My thought process was that one could update the CKM matrix and its dependent calculations right after editing the elements, but this makes sense as well.

MELA/interface/TUtil.hh
usarica: I think the defaults here should be motivated by the physics we are doing. Can you derive the default values of these parameters that correspond to the defaults in JHUGen?

MELA/interface/TEvtProb.hh
MohitS704: input is as follows:
nhel >= 0 -> that specific helicity
nhel = -1 -> sum over helicities
nhel < -1 -> invalid, goes to 0

usarica: Helicity can be positive, negative, or 0 (depending on what the spin quantum number is). What is the correspondence between MG numbering convention and physics?

MELA/interface/TUtil.hh
MohitS704: At the values I provided, these are the default values (sourced from `mod_parameters.F90` and the madgraph CKM function).

| Element | JHU | MG |
| ----------- | ----- | ------ |
| UD | 0.974285d0 | 0.974348875+0i |
| US | 0.225290d0 | 0.2265+0i |
| CS | 0.9734244d0 | 0.974348875+0i |
| CD | -0.225182d0 | -0.2265+0i |
| TB | 0.99912367d0 | 1 + 0i | 
| TS | -0.040920069d0 | -0.0405287775+0j |

The other terms in the matrix (at least for JHUGen) are sourced from these defaults, so I am running under the (probably valid) assumption that the author of SMEFTSIM did the calculations correctly.

Everything seems to agrees to a few decimal points.

MohitS704: You can find the values of the CKM elements at mod_parameters.F90 line 2773

usarica: I know what they are in mod_Parameters :)

A default of ckmlambda=-1, for instance, does not make physical sense. That would (should have) set Vud=Vcs=0.5, Vus=-Vcd = -1 if you use the parametrization in PDG; experimentally measured values are much more different. Leaving aside the fact that it is quite a bad convention to keep -1 as a replacement for defaults in this case, it is also not yet clear whether the PDG parametrization is used.

In the interest of making sure that ME calculations agree as much as possible, I would suggest to find the defaults of these parameters that correspond to what is in JHUGen. Complex CKM values are nice to have, but they create unnecessary differences between MEs that have little physical/experimental importance in Higgs physics.

MohitS704: Right - I set the defaults to what gets you the values in the table. The latest PR has this update.

MohitS704: Sorry, latest commit I mean.

usarica: Ok, almost there then. I suggest setting ckmeta=0 by default to keep the MEs as closely as possible. That should ensure Vtd and Vub, which you don't list above, are real. Their values are supposed to be small, <0.01 in absolute value, so their contribution to the full process is negligible.

MELA/interface/TEvtProb.hh
MohitS704: I looked deeper into this, and the helicity specification seems to be indexing all the different helicity combinations that are possible.

it looks something like `DATA (NHEL(I,  23),I=1,6) /-1, 1, 1, 1,-1,-1/` (2 gluons and 4 leptons), and is dependent upon the final state. I'm honestly not entirely sure what we would do to ensure that people know what helicity corresponds to what value. 

I'm hesitant to remove the helicity calculation options altogether, as it may be useful to someone in the future. Perhaps we can make an appendix to some sort of manual that details all the possible combinations?

MELA/interface/TUtil.hh
MohitS704: I set cmkmeta to be 0

MohitS704:
I think all that's left is the matter of the helicities - we can discuss what to do there.

MELA/interface/TEvtProb.hh
usarica: Is there a way to look at the LHE record of helicities when you generate a few events with different nhel options?

MohitS704: I'm not sure what you mean by this - nhel isn't a generation option. 

There is a list of all the possible helicitiies (however many there are - in ggF for this case there are 64). They are numbered from 1 to 64, each of them corresponding to a different setup like shown above (with nhel=23). When nhel=-1, all the possible contributions are summed over to make something equivalent to what MELA in MCFM and JHUGen does. When a specific nhel is selected, it only looks at the contributions from that specific final state of helicities.

I've already asked on the MG@NLO forums how to access the helicity configuration for each "index" without looking at the code, if that's possible.

MohitS704: I'm not sure what you mean by this - nhel isn't a generation option. 

There is a list of all the possible helicitiies (however many there are - in ggF for this case there are 64). They are numbered from 1 to 64, each of them corresponding to a different setup like shown above (with nhel=23). When nhel=-1, all the possible contributions are summed over to make something equivalent to what MELA in MCFM and JHUGen does. When a specific nhel is selected, it only looks at the contributions from that specific final state of helicities.

I've already asked on the MG@NLO forums how to access the helicity configuration for each "index" without looking at the code, if that's possible.

MohitS704: You can see the question here:
https://answers.launchpad.net/mg5amcnlo/+question/817453

usarica: Thanks, Mohit. I suspect the different combinations are just the signs of the helicities of the 6 particles.

Assuming this is the case, let's first check which sign corresponds to left- vs. right-handedness while MG experts take the time to respond. Are you able to modify Z couplings? If so, can you set the right- or left-handed Z->ll couplings to 0 manually and check which nhel configurations give a nonzero ME in an arbitrary set of Higgs couplings?

MohitS704: I have answers from MG experts on this topic. So, I have confirmed that we have access to the information contained within the helicities, and a way to find them. I also have information from the original [HELAS paper](https://inspirehep.net/literature/336604) that shows the helicity relationship. The image is for fermions, but the point still stands.

image


Helicities are also recorded in LHE files for things generated with Madgraph - but JHUGen/MCFM does not do this - we just report all of them as the same number.

My main point (and I'm going to go away from my original opinion here) is that I don't actually think we need the helicity stuff anyways - it seems irrelevant for our use cases (as seen by the fact that we ignore it completely when we do our own Monte Carlo!).

This is my opinion as of right now - and I think we should move forwards with a simple -1 in the nhel category always, removing the option from computeP. In the future, if necessary, we can integrate and revisit it, but I have doubts we ever will.

usarica: > Helicities are also recorded in LHE files for things generated with Madgraph - but JHUGen/MCFM does not do this - we just report all of them as the same number.

Correct. This is why I was asking this information in the first place, and to understand the exact relationship between the LHE record and the MG convention.

I think in general having access to exclusive helicity states is valuable; this is something I hoped could also be done for JHUGen MEs in the future as well (if you think about a massive parallelization scheme using GPUs etc., there is absolutely no reason why the Fortran code should be doing the summation over different helicity configurations - that is a lower-level job that is suitable for GPUs, with interface through C++ code - not Fortran!). It is a matter of preparing the input record for MEs rather than a ME property itself - if the function that calls the ME knows how to recognize the helicity information, it should be able to pick it up without requiring the user to do anything extra in the function call. In that sense, it might be more general to implement helicity information directly into the MELAParticle class, e.g., with an additional char (I don't think any of our particles will have |m_s/s|>127).

I would leave it to you to decide whether you would like to incorporate this information in such a way, or postpone for another time in case you have other time considerations.

MohitS704: I think that for now, seeing no imminent use case, I will default to a matrix element that is the result of summing over many of them.

MohitS704:
> I've added a line and another URL to `retrieve.csh` to allow for the user to wget the file. Once this PR is done (and the structure of libSMEFTsim.so finalized), we can upload the file PR.

A PR shouldn't be needed for this as we will simple add a file to the relevant architecture folder.

MohitS704:
@usarica I'm pretty sure everything is done now

usarica (approved):
I think the changes are now ok. I might also suggest to add ME checks through the testME_v2 script, but it is not a showstopper.

gritsan (approved):

PR #69, opened 2025-11-27
Cloudflare on jhu.edu now forbids wget, which breaks the NNPDF download script

namapane:
Some security update in spin.pha.jhu.edu now blocks downloading files with wget, which breaks the wget of NNPDF30_lo_as_0130.LHgrid.tgz in MELA/downloadNNPDF.sh, which is called in the checkout script.
(note that the file is still accessibe from a browser; the server only bloks wget, curl etc.)

Wgetting from the original distribution (https://pcteserver.mi.infn.it/~nnpdf/nnpdf30/NNPDF30_lo_as_0130.LHgrid.tgz) now works (even if a comment in the script stated otherwise; which may have been due to a temporary issue), so for the time being I propose to get it from there. 
I checked that the untarred file is identical.


gritsan:
I have been in contact with IT support at JHU and the issue has been resolved. wget from https://spin.pha.jhu.edu/ should now work. They did not tell me what was wrong, but whatever it is, it is fixed, I have just checked from lxplus.

namapane:
Great! Thanks for your prompt intervention. I confirm everything works now.

